PR #5, opened 2025-04-30
Initial cleanup

MaximilianSoerenPollak:
Cleaned some things.

Moved BUILD files into their own folders. 
Enabled cleaner conf.py
Moved docs.bzl 

There is still more Cleanup to be done, especially in the code itself (deleting all the comments / debug things etc.) but those can be done in future PR's that are focused on that alone. THis is already big enough

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -20,13 +20,9 @@ load("@rules_python//python:pip.bzl", "compile_pip_requirements")
 load("@rules_python//sphinxdocs:sphinx.bzl", "sphinx_build_binary")
 load("@score_dash_license_checker//:dash.bzl", "dash_license_checker")
 
-
-
 # These are only exported because they're passed as files to the //docs.bzl
 # macros, and thus must be visible to other packages. They should only be
 # referenced by the //docs.bzl macros.
-
-
 
 exports_files(
     [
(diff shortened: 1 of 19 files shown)


src/BUILD
```diff
@@ -98,6 +89,19 @@ py_library(
...
+    srcs = glob(
+        ["*/**"],
+        exclude = [
+            "**/test/**",
+            "**/tests/**",
+            "**/__pycache__/**",
+        ],
+    ),
+    visibility = ["//visibility:public"],
+)
```
MaximilianSoerenPollak, line 103: For the source_code_linker to scan.

src/incremental.py
```diff
@@ -46,20 +46,20 @@ def transform_env_str_to_dict(external_needs_source: str) -> list[dict[str, str]
...
-        l = {a[0]: a[1] for a in b}
-        if l:
-            l_dict.append(l)
-    return l_dict
+    for inner_dict in dict_list:
+        kv_splits = [kv.split(":", 1) for kv in inner_dict if len(inner_dict) > 1]
+        single_dict = {key_value[0]: key_value[1] for key_value in kv_splits}
+        if single_dict:
+            transformed_dicts.append(single_dict)
+    return transformed_dicts
```
MaximilianSoerenPollak, line 62: Renamed in order to make it a bit more readable and nicer.
Otherwise when we come back and need to change this it will be so bad

src/BUILD
AlexanderLanin: Is that a missing code comment? 😆

AlexanderLanin (approved):

src/BUILD
MaximilianSoerenPollak: Just an explenation. 

But yes it probably should be made a comment to add context.

PR #6, opened 2025-04-30
Adding initial README

MaximilianSoerenPollak:


AlexanderLanin (approved):
I guess the import problem applies to many more files and not only score_layout?!

PR #20, opened 2025-05-09
Fixing bug that broke 'docs_needs'

MaximilianSoerenPollak:
The 'dict' default value broken the parsing we had inside 'metamodel/__init__.py' to parse that value when  docs_needs  was build. 

This changes the default value to the expected one (list).

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -27,4 +27,5 @@ jobs:
         uses: bazel-contrib/setup-bazel@0.9.1
       - name: Run formatting checks
         run: |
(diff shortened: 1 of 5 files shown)


MaximilianSoerenPollak:
This isn't perfect / ideal longterm, to just 'run' the  ide_support beforehand. 


But after trying for 2 days I think I tried everything that came to mind and also I found online, and I think for now it's okay-ish. 
We should find a better more 'baseliks' way of doing this, for sure though, in the future.

AlexanderLanin (approved):

PR #21, opened 2025-05-12
Add default values for needs.json & bugfixes

MaximilianSoerenPollak:


MaximilianSoerenPollak:
Saw that the 'venv' was missing all requirements (which it should have in addition to all local modules)

AlexanderLanin (approved):

PR #16, opened 2025-05-09
docs-as-code: Add CICD workflow to the repository

Aymen-Soussi-01:
Add copyright, format, gitlint, licence_check and test workflows to the cicd of the repository.

Also-by: Aymen Soussi aymen.soussi@expleogroup.com

PR #34, opened 2025-05-20
docs: extened docs target

dcalavrezo-qorix:
added target for building github archive

Addresses: #16


MaximilianSoerenPollak (approved):
We most likely need to adapt the CI/CD targets once we actually build the documentation.

PR #37, opened 2025-05-21
Fixed wrong check activation

MaximilianSoerenPollak:
Check found False positives due to 'process' being moved and loosing it's prefix.

I would like some feedback to this, I don't 100% like the solution as it's a rather specific solution with potentially some unintended side effects, though I couldn't think of anything else currently. 

I did try moving the 'process_description' around so that 'process' is again in the document name, this however proofed to not be working, though I might have screwed this up somehow and not realized.


masc2023 (approved):

dcalavrezo-qorix (approved):

masc2023 (approved):

MaximilianSoerenPollak:
Also locally I have checked this against 'score' and score seems unaffected by this.

PR #58, opened 2025-05-28
Increase version for release

MaximilianSoerenPollak:
~~### ⚠️ Blocked until python_basics 0.3.2 release => https://github.com/eclipse-score/bazel_registry/pull/50~~

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -13,7 +13,7 @@
 
 module(
     name = "score_docs_as_code",
-    version = "0.2.7",
+    version = "0.3.0",
     compatibility_level = 0,
 )
 


MaximilianSoerenPollak:
It does seem that this works `test --build_tests_only` [docs](https://bazel.build/versions/7.4.0/docs/user-manual)

It does not build the documentation of sphinx. 

I tried the 'tag' filter here and it seems that didn't work. 

Eventhough they specified that they fixed it in this [ticket here](https://github.com/bazel-contrib/rules_python/issues/2645) we still get that issue currently. 

So I think this seems like the best solution right now?

MaximilianSoerenPollak:
> Currently `docs-as-code` is using `score_python_basics` 0.3.1, which uses prehistoric `rules_python` 1.0.0
> 
> This is already "fixed" in `tooling` release v0.6.6!
> 
> But there is no new `score_python_basics` release in `bazel_registry` based on the `tooling` v0.6.6
> 
> --> all this releasing and versioning is too complex at the moment!! It's super confusing when _tooling releases do not lead to releases of tools._

I have upgraded the python_tooling (locally) to 3.2 (so the version where it also uses python 1.4.0. 

But I still run into this issue. 

Will try to experiment further seeing if this is just a 'me' problem.

AlexanderLanin (approved):

PR #61, opened 2025-06-03
Fixed exeternal prefix cleaning

MaximilianSoerenPollak:
Fixing external prefix cleaning
Adding tests via RST files
Changing metamodel to allow for tests

### DIFF
--- a/src/extensions/score_metamodel/checks/check_options.py
+++ b/src/extensions/score_metamodel/checks/check_options.py
@@ -53,10 +53,10 @@ def validate_fields(
     :param field_type: A string indicating the field type ('option' or 'link').
     """
 
-    def remove_prefix(values: list[str], prefixes: list[str]) -> list[str]:
+    def remove_prefix(word: str, prefixes: list[str]) -> str:
         # Memory and allocation wise better to use a generator here.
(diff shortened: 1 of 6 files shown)


dcalavrezo-qorix (approved):

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -77,10 +77,10 @@ def remove_prefix(values: list[str], prefixes: list[str]) -> list[str]:
             values = [str(raw_value)]
 
         # The filter ensures that the function is only called when needed.
-        if field_type == "link" and allowed_prefixes:
-            values = remove_prefix(values, allowed_prefixes)
 
         for value in values:
+            if field_type == "link" and allowed_prefixes:
+                value = remove_prefix(value, allowed_prefixes)
```
MaximilianSoerenPollak, line 83: Issue was that if a list was passed into it, instead of removing all valued prefixes it just did the first value.

AlexanderLanin (approved):

src/extensions/score_metamodel/checks/check_options.py
AlexanderLanin: Should have been caught by basedpyright?!

PR #69, opened 2025-06-04
Update metamodel with

masc2023:
trustable framework needs,
https://github.com/eclipse-score/process_description/pull/27

add security tag to the document need

Resolves: https://github.com/eclipse-score/score/issues/947


MaximilianSoerenPollak (changes requested):
Should bring the id regex in line with the rest. 

Also running: 

`bazel run //src:ide_support`
`bazel run //src:format.fix` 

Should fix the formatting errors

Issue #53, opened 2025-05-27
Bug: Version selector URL is hardcoded


MaximilianSoerenPollak:
Currently the 'version' selector inside the Docs is hardcoded to the 'score_platform' repo. 

[See here](https://github.com/eclipse-score/docs-as-code/blob/7bfafa46956a04892f9d86ea067dbafebe526147/src/extensions/score_layout/html_options.py#L60)

This should not be the case. It should always take the current repo as basis.

dcalavrezo-qorix:
![Image](https://github.com/user-attachments/assets/f31ed85f-2bc9-47c0-acf0-9deff06cd509)

nicu1989:
did a test PR after merge: https://github.com/eclipse-score/docs-as-code/pull/72

PR #70, opened 2025-06-04
docs: Add user and repo as parameters to incremental

nicu1989:
Switch to incremental build that now accepts github_user and github_repo as parameters.

Works only after merging: https://github.com/eclipse-score/cicd-workflows/pull/23

Ex: bazel run //docs:incremental_latest -- --github_user=my-handle --github_repo=my-repo

Fork preview: https://github.com/nicu1989/docs-as-code/pull/4

Related: #53

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -13,7 +13,7 @@
 
 module(
     name = "score_docs_as_code",
-    version = "0.3.1",
+    version = "0.3.2",
     compatibility_level = 0,
 )
 


.github/workflows/docs.yml
```diff
@@ -38,5 +38,5 @@ jobs:
       id-token: write
 
     with:
-      bazel-target: "//docs:github_pages__release"
+      bazel-target: "//docs:incremental_release -- --github_user=nicu1989 --github_repo=docs-as-code"
```
AlexanderLanin, line 41:
```suggestion
      bazel-target: "//docs:incremental_release -- --github_user=$GITHUB_REPOSITORY_OWNER --github_repo=$GITHUB_REPOSITORY"
```

But I'm not sure those are available within `with`. Maybe it's:

```suggestion
      bazel-target: "//docs:incremental_release -- --github_user=${{env.GITHUB_REPOSITORY_OWNER} --env.github_repo=${{GITHUB_REPOSITORY}"
```

MaximilianSoerenPollak, line 41: if this is now used, could we not then delete all the 'github pages' stuff and make 'docs' simpler? 

Plus I think we also need to change the 'deployment' workflow as well, as we need to manualy tar it now & change the path to it, if I remember right.

nicu1989: Workflow is updated in this PR: https://github.com/eclipse-score/cicd-workflows/pull/23

.github/workflows/docs.yml
```diff
@@ -38,5 +38,5 @@ jobs:
       id-token: write
 
     with:
-      bazel-target: "//docs:github_pages__release"
+      bazel-target: "//docs:incremental_release -- --github_user=${{ github.repository_owner }} --github_repo=${{ github.event.repository.name }}"
```
AlexanderLanin, line 41: The parameters belong into cicd-workflows, don't they? As they are generic, we don't need to force users to write them out.

PR #76, opened 2025-06-04
Change hover color for visited links in the score grid

4og:
This sets the color of visited links when hovering them. The current color matches the grid background making the links look disappeared.

image


MaximilianSoerenPollak (approved):
Thanks for this. Will for sure make it easier to use :)

Issue #77, opened 2025-06-05
Bug: Filtering out external needs before checks autofails some checks


MaximilianSoerenPollak:
Due to our 'external filter' of needs that we have before our checks we currently only pass 'local' needs into the checks. 

This works fine as is, but *not* when you have a requirement that needs it's parents checked (for example for safety) that 'satisfies' or otherwise links to an EXTERNAL need.

Then suddenly the external need can not be found therefore the check fails. 

This should not happen.

PR #78, opened 2025-06-05
Bugfix: External needs graph checks & misc fixes

MaximilianSoerenPollak:
Our checks have filtered out 'external needs' completely, this lead to a bug that if a local need referenced an external need & that local need should be checked via a graph check, it will not find the linked need to check, as it is external and filtered. 

This PR addresses this bug behavior. 

--- 

PR addresses: 

- Graph checks now work correctly with local needs that link to external needs => changed behavior of filtering
- Source code linker not finding things => broke template strings until we have tool requirements.
- Print statements in tests were left uncommented  => commented them out
- Fixing links that no longer are in SCORE => change the linked need

fixes: #77

### DIFF
--- a/docs/example/index.rst
+++ b/docs/example/index.rst
@@ -30,7 +30,7 @@ This is a rendered example of the 'examples/linking-both' folder using the `docs
 
    Some content to make sure we also can render this
    This is a link to an external need inside the 'score' documentation.
-   :need:`SCORE_gd_req__req__attr_safety`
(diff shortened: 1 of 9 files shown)


dcalavrezo-qorix (approved):

MaximilianSoerenPollak:
I have split the commits here into groups. So each commit only touches the part of the msg, I think this might be easier to read then, if you go through bit by bit. 

I do kind of like this

AlexanderLanin (approved):

Issue #89, opened 2025-06-17
try setting up github cache for accelerating docs build


AlexanderLanin:
our docs:incremental build (see current workflows) is taking quite a long time.

Probably we can improve that by leveraging github cache abilities.

* see https://github.com/actions/cache
* figure out what exactly needs to be cached (e.g. entire _build folder, sub piece of it, etc)
* enable and configure caching in our workflows
  * only for PR builds? Just to make sure the main-builds are really 100% clean?!

PR #95, opened 2025-06-24
Add caching to the workflow

Aymen-Soussi-01:
As the experiment was done, I succeeded in reducing the build time after integrating the cache for different targets( bazel and python dependencies and rst files) into 2 workflows(format and test) from 5min 11s to 1min 48s. Note that the first workflow was a bit too long as it ran and stored the cache for the first time, but the normal runtime is 3min 40s, so we almost reduced it by half 
I should mention that I couldn't currently integrate the caching into more workflows as the rest of them integrate reusable workflows from other repositories and we should put the cache into the original reusable workflow repositories if we want to integrate the cache for them.

close: https://github.com/eclipse-score/docs-as-code/issues/89

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -48,6 +48,8 @@ use_repo(python)
 ###############################################################################
 pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
 pip.parse(
+    envsubst = ["PIP_INDEX_URL"],
+    extra_pip_args = ["--index-url=${PIP_INDEX_URL:-https://pypi.org/simple/}"],
     hub_name = "pip_process",
     python_version = PYTHON_VERSION,
     requirements_lock = "//src:requirements.txt",


AlexanderLanin:
I guess that's a valid solution until we have bazel remote caching. And indeed it saves ~2 minutes per run, which is 80-90% of time when docs themselves are trivial as in this repo.

However it's only applicable when docs is the only thing the workflow is building, as cache would become huge otherwise.

The hash string is missing some files?! Maybe everything in src? MODULE.bazel? Top level BUILD file? All bzl files? etc

Bazel has a disk_cache parameter, did you look into that?

And last but not least in cicd-workflows we have this action, why does it not work as-is?
```
      - name: Setup Bazel with shared caching
        uses: bazel-contrib/setup-bazel@0.14.0
        with:
          disk-cache: true
          repository-cache: true
          bazelisk-cache: true
```

Aymen-Soussi-01:
You're right — Bazel remote caching would be the ideal long-term solution. For now, this local cache setup helps save around 2 minutes per run, which is especially helpful when the docs are the only thing being built.
 I’ve now improved the cache key to include more relevant files: 
- All .bazel, .bzl, and BUILD files
-  The MODULE.bazel file
- Everything under src/, including Python files and requirements.txt

I also considered bazelisk's built-in caching (disk_cache, repository_cache)

I even used the latest version of setup-bazel, which is now 0.15.0

PR #103, opened 2025-06-25
Add support to inject other pip index

simon-d-bmw:
Prior to this change, python dependencies were only downloaded from the index
defined in requirements.txt. This is a limitation of rules_py, which usually uses
the bazel downloader for python dependencies. However, if a locked
requirements.txt is provided, it directly uses pip to download patches
from the index specified in the file.

Allow specifying custom index URLs through an environment variable.


AlexanderLanin:
I'll merge for now, as to not block this. But let's address those questions.

PR #114, opened 2025-06-30
checks: bugfix id contains feature

hoe-jo:
- separate feature in single parts
- search for each part individually

### DIFF
--- a/src/extensions/score_metamodel/checks/id_contains_feature.py
+++ b/src/extensions/score_metamodel/checks/id_contains_feature.py
@@ -37,6 +37,7 @@ def id_contains_feature(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
 
     # Get the part of the string after the first two underscores: the path
     feature = parts[1]
+    featureparts = feature.split("_")
 
     dir_docname = os.path.dirname(str(need.get("docname", "")))
 
@@ -45,7 +46,12 @@ def id_contains_feature(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
     # NOTE: This does not match the process requirements
     docname = dir_docname if dir_docname else need.get("docname", "")
 
-    if feature not in docname:
+    found = False
+    for featurepart in featureparts:
+        if featurepart in docname:
+            found = True
+
+    if not found:
         log.warning_for_option(
             need, "id", f"Feature '{feature}' not in path '{docname}'."
         )


src/extensions/score_metamodel/checks/id_contains_feature.py
```diff
@@ -45,7 +46,12 @@ def id_contains_feature(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
     # NOTE: This does not match the process requirements
     docname = dir_docname if dir_docname else need.get("docname", "")
 
-    if feature not in docname:
+    found = False
+    for featurepart in featureparts:
+        if featurepart in docname:
+            found = True
```
MaximilianSoerenPollak, line 52: A suggestion that avoids a for loop. Not sure if this is more readable, but would be faster. 
I'm okay with either. @AlexanderLanin what's your opinion? 

```suggestion
found = any(featurepart.lower() in docname.lower() for featurepart in featureparts)
```
We should  make it case insensitive, though.

AlexanderLanin:
I'm not so sure of the intention here. This contradicts requirements in my opinion.

AlexanderLanin:
It would certainly become clearer with documentation and/or tests ;-)

hoe-jo:
Minutes from Operational Support Meeting:
- Implement first letter abbreviation
- Match whole substring in current implementation

src/extensions/score_metamodel/checks/id_contains_feature.py
```diff
@@ -37,6 +38,7 @@ def id_contains_feature(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
 
     # Get the part of the string after the first two underscores: the path
     feature = parts[1]
+    featureparts = re.split(r"_-", feature)
```
MaximilianSoerenPollak, line 41:
```suggestion
    featureparts = re.split(r"[_-]", feature)
```
I think it needs to be this, otherwise it will look for '_-' instead of '-' or '_'

src/extensions/score_metamodel/checks/id_contains_feature.py
```diff
@@ -45,7 +47,22 @@ def id_contains_feature(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
...
 
-    if feature not in docname:
+    # allow if any feature part is contained in UID
+    foundfeatpart = any(
+        featurepart.lower() in docname.lower()
+        for featurepart in featureparts
+        if featureparts
+    )
+
+    # allow abbreviation of the feature
```
MaximilianSoerenPollak, line 57: Just a note: 

This makes initials of number stuff a bit unideal, but I don't mind that. 

e.g. aspice 40 => a4

MaximilianSoerenPollak:
Rebased to resolve merge conflicts.

PR #119, opened 2025-07-01
Add support for injecting dependencies

simon-d-bmw:
Users of docs-as-code can now provide additional dependencies
to the docs, and ide_support macros, to support use cases like
project-specific Pyhton packages.


AlexanderLanin:
Adding to venv is fine. But I'm not sure about the other one. I don't think we want to allow arbitrary dependencies to docs. What's the use case here?

AlexanderLanin (approved):

PR #120, opened 2025-07-01
docs: Fix registration of HeaderService

nicu1989:
Esbonio crashes due to score_header_service extension being registered when trying to build for the second time.

Related: https://github.com/eclipse-score/score/issues/1117


AlexanderLanin:
Thanks for that analysis!

If this PR doesn't work, just disable header service completely! We don't use it at the moment, or anytime soon.

AlexanderLanin (approved):

PR #124, opened 2025-07-04
Remove empty globs

simon-d-bmw:
Add support for bazel setting incompatible_disallow_empty_glob to detect
unintentional empty globs.

While doing that, I detected that there are some filegroups with empty globs:

* score_extension_files, replace its usages with docs_as_code_py_modules.
* test_rst_files: test content already covered by glob in score_metamodel_tests

In both cases, the globs were targeting subdirectories which are themselves bazel packages.
Therefore the glob does not match them.

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -83,7 +83,7 @@ def docs(source_files_to_scan_for_needs_links = None, source_dir = "docs", conf_
         sphinx_build_binary(
             name = "sphinx_build" + suffix,
             visibility = ["//visibility:public"],
-            data = ["@score_docs_as_code//src:docs_assets", "@score_docs_as_code//src:score_extension_files"] + external_needs_deps,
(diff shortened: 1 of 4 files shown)


MaximilianSoerenPollak:
I have not heard yet. So I would say we can merge this. 

@simon-d-bmw Could you rebase this, I will then approve & merge this. 

Thanks again for the effort.

src/extensions/score_source_code_linker/BUILD
```diff
@@ -30,8 +30,7 @@ score_py_pytest(
     name = "score_source_code_linker_test",
     size = "small",
     srcs = glob([
-        "tests/**/*.py",
-        "test/**/*.json",
```
simon-d-bmw, line 34: Not sure about the intention here, but the macro only accepty python files here. That's why I removed this json glob.

PR #129, opened 2025-07-09
WIP: Consumer tests

MaximilianSoerenPollak:
Developing the consumer tests here for docs-as-code. 

Just want to gather some feedback here on the general appraoch etc. 

--- 

UNFINISHED therefore done as draft.


src/tests/test_consumer.py
```diff
@@ -0,0 +1,571 @@
...
+        module_git_override = replace_bazel_dep_with_git_override(
+            module_orig, current_hash, gh_url
+        )
+        with open("MODULE.bazel", "w") as f:
+            f.write(module_local_override)
+        for cmd in repo.commands:
+            # Running through all 'cmds' specified with the local override
+            print_running_cmd(repo.name, cmd, "LOCAL OVERRIDE")
+
+            out = subprocess.run(cmd.split(), capture_output=True, text=True)
```
nicu1989, line 449: why not check the subprocess for errors also here? with check=True. I feel like if some future maintainer “optimises” the processing of BR and forgets to look at returncode failures silently become successes.

MaximilianSoerenPollak: The issue is, that if I do 'check=True' here, if bazel exits with a non 0 exit code (which it does when sphinx errors for example) it just stops execution immediatly there. 
At least to my testing. That's why I thought it better to not do this here.

nicu1989: understood

src/tests/test_consumer.py
```diff
@@ -0,0 +1,571 @@
...
+
+
+def replace_bazel_dep_with_local_override(module_content: str) -> str:
+    """ """
+
+    # Pattern to match the bazel_dep line
+    pattern = rf'bazel_dep\(name = "score_docs_as_code", version = "[^"]+"\)'
+
+    # Replacement with local_path_override
+    replacement = f"""bazel_dep(name = "score_docs_as_code", version = "0.4.2")
```
AlexanderLanin, line 195: Force failure without override:
```suggestion
    replacement = f"""bazel_dep(name = "score_docs_as_code", version = "0.0.0")
```

src/tests/test_consumer.py
```diff
@@ -0,0 +1,571 @@
...
+    commit = "{git_hash}"
+)'''
+
+    modified_content = re.sub(pattern, replacement, module_content)
+
+    return modified_content
+
+
+def parse_bazel_output(BR: BuildOutput) -> BuildOutput:
+    # HINT: All errors from sphinx / bazel commands are put into the stderr for some reason.
```
AlexanderLanin, line 224: because stderr is for errors... 😆

src/tests/test_consumer.py
```diff
@@ -0,0 +1,571 @@
...
+
+    # Getting GH_URL & current hash to enable git override
+    gh_url = get_github_base_url(git_root)
+    current_hash = get_current_git_commit(curr_path)
+
+    # Symlink local docs-as-code => sphinx tmp
+    docs_as_code_dest = sphinx_base_dir / "docs_as_code"
+    docs_as_code_dest.symlink_to(git_root)
+
+    # Calculating / Checking if the metamodel.yaml has changed, to determin if [score_metamodel] errors should be disablede
```
AlexanderLanin, line 422: not sure why we would want that

MaximilianSoerenPollak:
@AlexanderLanin Where should this test file be located btw? Not really sure if this is a good place for it, where I put it currently.

src/tests/test_consumer.py
MaximilianSoerenPollak: I'm guessing my brain was on break when I wrote this :D

Aymen-Soussi-01:
Don't forget to correct the formatting check

AlexanderLanin (approved):

PR #131, opened 2025-07-09
Bugfix: Attribute Safety Analysis

PandaeDo:
Bugfix for attributes linkage of DFA to static diagrams and changed mitigates to optional link if the violation doesn't apply

### DIFF
--- a/src/extensions/score_metamodel/metamodel.yaml
+++ b/src/extensions/score_metamodel/metamodel.yaml
@@ -512,7 +512,7 @@ needs_types:
       id: ^feat_plat_saf_dfa__[0-9a-z_]+$
       violation_id: ^.*$
       violation_cause: ^.*$
-      violation_applicable: ^(yes|no)$
+      sufficient: ^(yes|no)$
       status: ^(valid|invalid)$
     optional_options:
       sufficient: ^(yes|no)$


src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -531,7 +531,7 @@ needs_types:
       status: ^(valid|invalid)$
     mandatory_links:
       mitigates: ^(feat_req__.*|aou_req__.*)$
-      verifies: ^feat_arc_dyn__[0-9a-z_]*$
+      verifies: ^feat_arc_sta__[0-9a-z_]*$
```
MaximilianSoerenPollak, line 534: So for `feat_saf_dfa` mitigates is a mandatory link, not optional like above? 

Should it not be always the same?

MaximilianSoerenPollak: So this is now correct, as it is?  With the links being mandatory?

MaximilianSoerenPollak: Will you then change mandatory / optional things to be the same everywhere? 

I see sufficent being optional and the new option of `violation_applicable' etc.

PandaeDo: Changed back to sufficient

PR #134, opened 2025-07-11
Add cli help targets description.

nicu1989:
added CLI help updates


MaximilianSoerenPollak (approved):
👍  I think this looks like a good solution

PR #140, opened 2025-07-17
Updating Tool requirements

MaximilianSoerenPollak:
With updating and continuously clarifying process requirements, our tool requirements also need to change and be adapted.

As well as adding anything we might have missed so far.


MaximilianSoerenPollak:
These will be updated further once the process upgraded to new release in a week or so.

PR #157, opened 2025-07-23
Automatically Execute ide_support Command

opajonk:
Run the respective Bazel command on container start (but after the `postCreateCommand` of the container features!). Select the created venv as default. And fix a wrong path in the README.


MaximilianSoerenPollak (approved):

AlexanderLanin:
I think `postCreateCommand` is too late as it runs *after* caching. On the other hand `onCreateCommand` is executed before caching.

(Caching seems disabled, we'll need to check how to enable it here without admin access)

PR #167, opened 2025-07-24
Fixing consumer_test triggers & color output

MaximilianSoerenPollak:
Now adds the check to the 'check list' in the PR with the name of the triggerer. 
image

It also inside the CI now has colored output
image

### DIFF
--- a/src/tests/test_consumer.py
+++ b/src/tests/test_consumer.py
@@ -20,6 +20,7 @@ import logging
 from pathlib import Path
 from rich import print
 from rich.table import Table
+from rich.console import Console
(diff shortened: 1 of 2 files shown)


src/tests/test_consumer.py
```diff
@@ -49,6 +50,8 @@
 LOGGER = logging.getLogger(__name__)
 LOGGER.setLevel("DEBUG")
 
+console = Console(force_terminal=True if os.getenv("CI") else None, width=120)
```
AlexanderLanin, line 53: does rich pick this up automatically? Is Console a singleton?

PR #177, opened 2025-07-25
Improve consumer test

MaximilianSoerenPollak:
Improving consumer tests. 

- Enabled local caching via command line flag
- Enabled streaming output without parsing on `-vvv` to give faster feedback
- Added a README to explain some processes on how the script works locally & how to execute it in CI 
- Added some further debug printing statements.
- Added parsing & cleaning of ANSI color codes inside of CI.


MaximilianSoerenPollak:
@AlexanderLanin Please let me know if any further feature requests arise that this didn't cover.

src/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+def pytest_addoption(parser):
+    """Add custom command line options to pytest"""
+    parser.addoption(
+        "--repo-tests",
```
AlexanderLanin, line 16: Not sure why this has a tests suffix?

```suggestion
        "--repo",
```

src/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+        action="store",
+        default=None,
+        help="Comma separated string of ConsumerRepo's name tests to run",
+    )
+    parser.addoption(
+        "--keep-temp",
+        action="store_true",
+        default=False,
```
AlexanderLanin, line 24: Shouldn't incremental be the default?

src/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+def pytest_addoption(parser):
```
AlexanderLanin, line 13: this will fail horribly once we have other tests in `tests`?

MaximilianSoerenPollak: Wait? 

This seems to not have taken my latest commit? 
I have removed this and left it only in the conftest.py as this code block, doesn't have any effect in the file itself according to my testing. 

Weird oO. Let me double check my branch

MaximilianSoerenPollak: This i haven't tested. 
But it shouldn't, as far as I understand pytestconfig

src/tests/conftest.py
MaximilianSoerenPollak: It is now.

Issue #145, opened 2025-07-18
tool_req__docs_common_attr_id_scheme: ID Schema clarification


MaximilianSoerenPollak:
[tool_req__docs_common_attr_id_scheme](https://eclipse-score.github.io/docs-as-code/pr-140/product/requirements.html#tool_req__docs_common_attr_id_scheme) currently talks about this ID schema. 

[Code link](https://github.com/eclipse-score/docs-as-code/blob/main/src/extensions/score_metamodel/checks/attributes_format.py#L21-L71) It has a lot of exceptions, that are not clearly defined. 

We need to make sure these exceptions are correct and somehow document them. 

- [ ] Clarify Exceptions with Process Team
- [ ] If necessary change implementation

PR #186, opened 2025-07-29
Remove check id format and change it with regex check in the metamode…

Aymen-Soussi-01:
Remove check id format and change it with regex check in the metamodel and adapt test cases

close: https://github.com/eclipse-score/docs-as-code/issues/145


AlexanderLanin:
/consumer-test

(just testing consumer tests)

PR #202, opened 2025-08-11
Complete integration of the new linter from tooling into docs-as-code

Aymen-Soussi-01:
Integrate necessary modules related to the new Linter introduced here: https://github.com/eclipse-score/tooling/tree/main/tools
Add and run the sample.sh script successfully detects errors related to Linting and formatting.

Result: we found 129 errors, some of which are related to code line length. We chose not to correct them directly as rough commands cannot correct them, and we may make this in another PR.

Just let me know if you want to include the fix for the errors inside this PR .

close: https://github.com/eclipse-score/docs-as-code/issues/199


sample.sh
```diff
@@ -0,0 +1,36 @@
+# *******************************************************************************
```
MaximilianSoerenPollak, line 1: This file should be renamed. 
Should not be 'sample.sh'

sample.sh
```diff
@@ -0,0 +1,36 @@
...
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
```
MaximilianSoerenPollak, line 2: We might then also want those things as an optional check in a workflow. 
So we see what is bad / not working, but not yet enforce it yet.

Aymen-Soussi-01: So we need to make a new workflow to run the Linting instead of including the sample.sh

AlexanderLanin: But we still need a linter.sh? For checking locally?!

MODULE.bazel
```diff
@@ -96,3 +96,21 @@ bazel_dep(name = "score_dash_license_checker", version = "0.1.1")
 
 # docs dependency
 bazel_dep(name = "score_process", version = "1.1.0")
+
+# Add Linter
```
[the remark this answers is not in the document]
MaximilianSoerenPollak: This already is condensed. 

Otherwise it would be 6 lines per Linter I think. 
Do not quiet remember. 

Though there might be a way to condense it some more.

sample.sh
Aymen-Soussi-01: We can now use the same command used in the new linter workflow, "bash scripts/run-linters.sh" to run it locally

Aymen-Soussi-01: @MaximilianSoerenPollak  can you check the PR update?

AlexanderLanin (approved):
merge as is

PR #207, opened 2025-08-13
Adding helper_lib

MaximilianSoerenPollak:
Added helper lib that contains some often used functions

### DIFF
--- a/src/extensions/score_source_code_linker/tests/test_source_link.py
+++ b/src/extensions/score_source_code_linker/tests/test_source_link.py
@@ -255,7 +255,7 @@ def example_source_link_text_non_existent(sphinx_base_dir):
         {
             "TREQ_ID_200": [
                 NeedLink(
-                    file=Path(f"src/bad_implementation.py"),
+                    file=Path("src/bad_implementation.py"),
                     line=2,
                     tag="#" + " req-Id:",
                     need="TREQ_ID_200",
@@ -266,16 +266,16 @@ def example_source_link_text_non_existent(sphinx_base_dir):
     ]
 
 
-def make_source_link(ws_root: Path, needlinks):
+def make_source_link(needlinks):
     return ", ".join(
-        f"{get_github_link(ws_root, n)}<>{n.file}:{n.line}" for n in needlinks
+        f"{get_github_link(n)}<>{n.file}:{n.line}" for n in needlinks
     )
 
 
 def compare_json_files(file1: Path, golden_file: Path):
-    with open(file1, "r") as f1:
+    with open(file1) as f1:
         json1 = json.load(f1, object_hook=needlink_test_decoder)
-    with open(golden_file, "r") as f2:
+    with open(golden_file) as f2:
         json2 = json.load(f2, object_hook=needlink_test_decoder)
     assert len(json1) == len(json2), (
         f"{file1}'s lenth are not the same as in the golden file lenght. "
@@ -315,7 +315,7 @@ def test_source_link_integration_ok(
             assert f"TREQ_ID_{i}" in needs_data
             need_as_dict = cast(dict[str, object], needs_data[f"TREQ_ID_{i}"])
             expected_link = make_source_link(
-                ws_root, example_source_link_text_all_ok[f"TREQ_ID_{i}"]
+                example_source_link_text_all_ok[f"TREQ_ID_{i}"]
             )
             # extra_options are only available at runtime
             # Compare contents, regardless of order.
--- a/src/extensions/score_source_code_linker/tests/test_source_link.py
+++ b/src/extensions/score_source_code_linker/tests/test_source_link.py
@@ -26,10 +26,8 @@ from sphinx_needs.data import SphinxNeedsData
 from test_requirement_links import needlink_test_decoder
 
 from src.extensions.score_source_code_linker import get_github_base_url, get_github_link
-from src.extensions.score_source_code_linker.generate_source_code_links_json import (
-    find_ws_root,
-)
 from src.extensions.score_source_code_linker.needlinks import NeedLink
+from src.helper_lib import find_ws_root
 
 
 @pytest.fixture()
--- a/src/extensions/score_source_code_linker/tests/test_source_link.py
+++ b/src/extensions/score_source_code_linker/tests/test_source_link.py
@@ -267,9 +267,7 @@ def example_source_link_text_non_existent(sphinx_base_dir):
 
 
 def make_source_link(needlinks):
-    return ", ".join(
-        f"{get_github_link(n)}<>{n.file}:{n.line}" for n in needlinks
-    )
+    return ", ".join(f"{get_github_link(n)}<>{n.file}:{n.line}" for n in needlinks)
 
 
 def compare_json_files(file1: Path, golden_file: Path):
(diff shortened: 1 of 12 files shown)


src/helper_lib/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    #       certain execution types, like esbonio
+    #       from recousing forever if it can't find a .git
+    #       in the 'home' or 'root' folder.
+    recoursion_break = False
+    while not (git_root / ".git").exists():
+        git_root = git_root.parent
+        if git_root == Path("/"):
+            # fallback to cwd when building with python -m sphinx docs _build -T
+            if recoursion_break:
+                return None
```
AlexanderLanin, line 58: Is it possible to write that loop somehow differently? I still don't get it. And there is no fallback to cwd?

src/helper_lib/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    path = url.split(":")[1] if url.startswith("git@") else "/".join(url.split("/")[3:])
+    return path.replace(".git", "")
+
+
+def get_github_repo_info(git_root_cwd: Path) -> str:
+    """
+    Unlike other funcions, this function behaves the same for the following commands:
+
+    'bazel run' => str
+    'bazel build' => str
```
AlexanderLanin, line 92: bazel build should not have .git in the sandbox? And it could run remotely where there is definitely no .git.

src/helper_lib/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        repo = parse_remote_git_output(process.stdout.split("\n")[0])
+    assert repo != "", (
+        "Remote repository is not defined. Make sure you have a remote set. "
+        + "Check this via 'git remote -v'"
+    )
+    return repo
+
+
+def get_git_root(git_root: Path = Path()) -> Path:
```
AlexanderLanin, line 120: get_git_root vs find_git_root is unclear

src/helper_lib/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+    if git_root == Path():
+        passed_git_root = find_git_root()
+        if passed_git_root is None:
+            return Path()
+    else:
+        passed_git_root = git_root
+    return passed_git_root
+
+
+def get_github_base_url(git_root: Path = Path()) -> str:
```
AlexanderLanin, line 142:
```suggestion
def get_github_pages_url(git_root: Path = Path()) -> str:
```

src/helper_lib/__init__.py
MaximilianSoerenPollak: Funny you mention that. I too am struggling to remember why we had this. 
Was copied from the place we had those functions before. 

I know it makes testing possible, but there might be a nicer way or better way to name those.

src/helper_lib/__init__.py
AlexanderLanin: Consider [while + else](https://www.w3schools.com/python/gloss_python_while_else.asp). No idea whether that will make it better, but it sounds like a perfect fit.

src/extensions/score_source_code_linker/tests/test_source_link.py
```diff
@@ -266,16 +264,14 @@ def example_source_link_text_non_existent(sphinx_base_dir):
     ]
 
 
-def make_source_link(ws_root: Path, needlinks):
-    return ", ".join(
-        f"{get_github_link(ws_root, n)}<>{n.file}:{n.line}" for n in needlinks
-    )
+def make_source_link(needlinks):
+    return ", ".join(f"{get_github_link(n)}<>{n.file}:{n.line}" for n in needlinks)
```
MaximilianSoerenPollak, line 268: It seems that both formaters disagree on this. 

Ruff (via multitool) fixed it to be like the old style. 
Then our formatter fixed it again to be like the right.

src/extensions/score_source_code_linker/tests/test_source_link.py
```diff
@@ -266,16 +264,14 @@ def example_source_link_text_non_existent(sphinx_base_dir):
...
 
 def compare_json_files(file1: Path, golden_file: Path):
-    with open(file1, "r") as f1:
+    with open(file1) as f1:
         json1 = json.load(f1, object_hook=needlink_test_decoder)
-    with open(golden_file, "r") as f2:
+    with open(golden_file) as f2:
         json2 = json.load(f2, object_hook=needlink_test_decoder)
     assert len(json1) == len(json2), (
         f"{file1}'s lenth are not the same as in the golden file lenght. "
```
MaximilianSoerenPollak, line 277: Nit: 

```suggestion
        f"{file1}'s lenth is not the same as the golden file. "
```

src/helper_lib/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+    Example:
+        Input:  'origin git@github.com:MaximilianSoerenPollak/docs-as-code.git'
+        Output: 'MaximilianSoerenPollak/docs-as-code'
+    """
+    if len(str_line.split()) < 2:
+        LOGGER.warning(
+            f"Got wrong input line from 'get_github_repo_info'. Input: {str_line}. "
+            + "Expected example: 'origin git@github.com:user/repo.git'"
+        )
+        return ""
```
MaximilianSoerenPollak, line 69: Nit/Next PR: 

Look through this file to find out where we actually can throw exceptions etc. 
As this 'empty' return might actually do more harm than good.

src/helper_lib/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+    Execution context behavior:
+    - 'bazel run' => ✅ Correct GitHub URL
+    - 'bazel build' => ⚠️ Uses Path() fallback when git_root is None
+    - 'direct sphinx' => ✅ Correct GitHub URL
+
+    Returns:
+        GitHub URL in format 'https://github.com/user/repo'
+    """
+    passed_git_root = find_git_root()
+    print("THIS IS PASSED GIT ROOT IN GH BASE URL", passed_git_root)
```
MaximilianSoerenPollak, line 125: Delete this

src/helper_lib/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+            cwd=git_root,
+            capture_output=True,
+            check=True,
+        )
+        decoded_result = result.stdout.strip().decode()
+
+        assert all(c in "0123456789abcdef" for c in decoded_result)
+        return decoded_result
+    except Exception as e:
+        LOGGER.warning(f"Unexpected error: {git_root}", exc_info=e)
```
MaximilianSoerenPollak, line 158: Nit: 
```suggestion
        LOGGER.warning(f"Unexpected error while trying to get git_hash. Exceuted in: {git_root}", exc_info=e)
```

src/extensions/score_source_code_linker/tests/test_source_link.py
MaximilianSoerenPollak: the one that we run via format.fix `bazel run //src:format.fix`

MaximilianSoerenPollak:
Overall it looks alright, I did check and don't see any ruff errors in the helper_lib as far as I can tell. Just so I don't introduce new linter warnnings while we try to reduce old ones.

AlexanderLanin (approved):

Issue #201, opened 2025-08-11
Fix Source-Code-Linker strings in code


MaximilianSoerenPollak:
The SCL links in the Code are currently broken and point directly to the process. 
They should point to Tool_reqs instead.

AlexanderLanin:
why does it build without failing?

MaximilianSoerenPollak:
> why does it build without failing?

Links are intentionally broken so they do not get picked up by the SCL.  

instead of `req-Id: gd_req...`  it's something like `req- Id: ...`  or `req-qId: ...`

PR #208, opened 2025-08-13
Fix source code links strings in code

Aymen-Soussi-01:
- Fix source code links strings in code like "# req-qId", "# req- Id", "# req-#id" by the correct one "# req-Id"
- Replace old gd_req source code links by their correct related tool_req based on the Process Requirements Overview table

close: https://github.com/eclipse-score/docs-as-code/issues/201

### DIFF
--- a/src/extensions/score_metamodel/metamodel.yaml
+++ b/src/extensions/score_metamodel/metamodel.yaml
@@ -931,8 +931,8 @@ needs_extra_links:
 #   - condition: defines the condition that should be checked
 #     - [and / or / xor / not]
 ##############################################################
-# req- Id: gd_req__req_linkage_architecture
-# req- Id: gd_req__req_linkage_safety
+# req-Id: gd_req__req_linkage_architecture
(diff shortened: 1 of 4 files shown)


src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -931,8 +931,8 @@ needs_extra_links:
 #   - condition: defines the condition that should be checked
 #     - [and / or / xor / not]
 ##############################################################
-# req- Id: gd_req__req_linkage_architecture
-# req- Id: gd_req__req_linkage_safety
+# req-Id: tool_req__docs_dd_feature_flag
```
AlexanderLanin, line 934: feature_flag?

Aymen-Soussi-01: This is the related tool_req from the Process Requirements Overview table
Screenshot-alex

AlexanderLanin:
Let's simply remove the two complex ones, as we need to redo them anyway. and merge the rest as-is

PR #216, opened 2025-08-19
add puml-theme-score

kalu-an:
This PR is about

1. enable PlantUML in docs-as-code repo at all
2. introduce puml-theme-score.puml
3. extend src/extensions/score_layout/__init__.py to have `:config: score_config` available for `.. needs`

see also https://github.com/eclipse-score/score/pull/1416

### DIFF
--- a/src/extensions/score_draw_uml_funcs/helpers.py
+++ b/src/extensions/score_draw_uml_funcs/helpers.py
@@ -236,16 +236,18 @@ def get_real_interface_logical(
             logger.info(f"{logical_op}: not implemented by any logical operation")
             continue
 
-        real_ifops.extend(real_ifop) if real_ifop not in real_ifops else None
+        if real_ifop[0] not in real_ifops:
+            real_ifops.extend(real_ifop)
 
-        # Per definition a operation can only be included by one interface
+        # Per definition an operation can only be included by one interface
         real_iface = all_needs[real_ifop[0]].get("included_by")
 
         if not real_iface:
             logger.info(f"{real_ifop[0]}: not included in any interface")
             continue
 
-        real_ifaces.extend(real_iface) if real_iface not in real_ifaces else None
+        if real_iface[0] not in real_ifaces:
+            real_ifaces.extend(real_iface)
 
     return real_ifaces
 
--- a/src/extensions/score_header_service/header_service.py
+++ b/src/extensions/score_header_service/header_service.py
@@ -89,7 +89,7 @@ class HeaderService(BaseService):
         :param config: The configuration dictionary.
         :param kwargs: Additional keyword arguments.
         """
-        super(BaseService, self).__init__()
+        super().__init__()
 
     def request_from_directive(
         self, directive: SphinxDirective
--- a/src/extensions/score_metamodel/checks/attributes_format.py
+++ b/src/extensions/score_metamodel/checks/attributes_format.py
@@ -67,11 +67,11 @@ def check_id_length(app: Sphinx, need: NeedsInfoType, log: CheckLogger):
     that would be replaced by actually feature names.
     ---
     """
-    max_lenght = 45
+    max_length = 45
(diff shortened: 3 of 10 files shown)


src/assets/puml-theme-score.puml
```diff
@@ -0,0 +1,566 @@
...
+''
+'' conti theme is based on materia theme
+'' Author: Andreas Kaluza
+'' Copyright (c) 2012 by Andreas Kaluza
+''
+'' materia theme based off of the bootstrap theme of the same name
+'' https://bootswatch.com/materia/
+''
+'' Author: Brett Schwarz
+'' Copyright (c) 2019 by Brett Schwarz
```
MaximilianSoerenPollak, line 12: @AlexanderLanin I got no idea how this is to be handled with copyright here? 
does this file also need a normal cpright header?

kalu-an: @MaximilianSoerenPollak I´m not a license expert. Case I did something wrong let me know. I will change it according to your proposal.

kalu-an: @pahmann and  @AlexanderLanin since I do have a poor knowledge on dealing with license correctly I had chat with LLM, treat it like this.
 
What I understood is that we are allowed to modify an existing theme which has MIT with modification statement. A common practice is to introduce a directory for storing original Licensees.

Kindly ask for your assistance and suggestions. thanks in advance 

> A good example of a real project that keeps third-party licenses (including the original MIT license) in a separate folder is the Python packaging tool Setuptools.
> 
> How Setuptools Handles Third-Party Licenses
> Vendored third-party code (code not authored by the project but included in the codebase) is kept in subdirectories such as setuptools/_vendor/.
> 
> Licenses for those third-party packages are stored directly in the setuptools/_vendor/ subfolders, often in files named like LICENSE, LICENSE.txt, or something similar specific to each dependency.
> 
> Relevant license files must be included in the package: This not only fulfills the requirements of the MIT license (which requires retention of the license text in distributions) but is also common for Apache 2, BSD, and other open source licenses.

MaximilianSoerenPollak: > @pahmann and @AlexanderLanin since I do have a poor knowledge on dealing with license correctly I had chat with LLM, treat it like this.
> 
> What I understood is that we are allowed to modify an existing theme which has MIT with modification statement. A common practice is to introduce a directory for storing original Licensees.
> 
> Kindly ask for your assistance and suggestions. thanks in advance
> 
> > A good example of a real project that keeps third-party licenses (including the original MIT license) in a separate folder is the Python packaging tool Setuptools.
> > How Setuptools Handles Third-Party Licenses
> > Vendored third-party code (code not authored by the project but included in the codebase) is kept in subdirectories such as setuptools/_vendor/.
> > Licenses for those third-party packages are stored directly in the setuptools/_vendor/ subfolders, often in files named like LICENSE, LICENSE.txt, or something similar specific to each dependency.
> > Relevant license files must be included in the package: This not only fulfills the requirements of the MIT license (which requires retention of the license text in distributions) but is also common for Apache 2, BSD, and other open source licenses.

It should be fine, but I think we should wait for some answer from @pahmann , or maybe @masc2023 could chime in.

masc2023: I think we had a similar issue here with Trustable, compare https://github.com/eclipse-score/process_description/issues/37 and referred links, should be a general issue to be resolved

FScholPer: Please open an eclipse help desk ticket with request for clarification

kalu-an: @FScholPer done: [Incorporation / Usage of MIT License in Eclipse S-CORE](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6565)
Let me know if something is missing. thanks in advance

kalu-an:
Continue PR #232

PR #218, opened 2025-08-19
Add documentation check command and time tracking

AlexanderLanin:
Introduce a command to verify documentation (undocumented = experimental) and implement time tracking for the documentation build process.

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -86,6 +86,19 @@ def docs(source_dir = "docs", data = [], deps = []):
     )
 
     py_binary(
+        name = "docs_check",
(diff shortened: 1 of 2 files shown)


AlexanderLanin:
It's technically the same as needs_json, but it has a different target group:

* needs_json is to generate the needs.json
* docs_check is to quickly verify traceability (NEW)
* needs_json internally verifies traceability
* docs_check internally builds needs.json


In the future docs_check could generate a html one-pager.

MaximilianSoerenPollak (approved):

PR #270, opened 2025-10-07
Fix flaky test

### LINKED ISSUE #226: Bug: Test seems to behave flakey for unknown reason
The `test_source_link_integration_ok` test seems to behave sometimes flakey. Meaning it fails => rerun => it passes. 

I have not looked into it, but we should probably fix this, or at least mark the test as flakey? Though it shouldn't be?

Find the code here: [source code](https://github.com/eclipse-score/docs-as-code/blob/main/src/extensions/score_source_code_linker/tests/test_source_code_link_integration.py#L471)
Find the output when it fails here:
```
FAILED ../score_tooling+/python_basics/score_pytest::test_source_link_integration_ok - AssertionError: assert {'d', '5', 'p', '3', ':', 'c', 't', '6', 'h', ' ', 'l', 'L', '<', 'a', 'g', 'n', '>', ',', 'b', '2', '7', '.', 's', '9', 'm', 'f', 'e', '1', 'i', '4', 'r', 'u', '#', '0', 'o', '_', '/', 'y'} == {'d', '5', 'p', '3', ':', 'c', 't', '6', 'h', ' ', 'l', 'L', '<', 'a', 'g', 'n', '>', ',', 'b', '2', '7', '.', 's', '9', 'm', 'f', 'e', '1', 'i', '4', 'r', 'u', '8', '#', '0', 'o', '_', '/', 'y'}
  
  Extra items in the right set:
  '8'
  
  Full diff:
    {
        ' ',
        '#',
        ',',
        '.',
        '/',
        '0',
        '1',
        '2',
        '3',
        '4',
        '5',
        '6',
        '7',
  -     '8',
        '9',
        ':',
        '<',
        '>',
        'L',
        '_',
        'a',
        'b',
        'c',
        'd',
        'e',
        'f',
        'g',
        'h',
        'i',
        'l',
        'm',
        'n',
        'o',
        'p',
        'r',
        's',
        't',
        'u',
        'y',
    }
```

a-zw:
## 📌 Description

Enables a test previously marked as flaky and makes it stable.

Closes #226 

## 🚨 Impact Analysis

### DIFF
--- a/src/extensions/score_source_code_linker/tests/test_source_code_link_integration.py
+++ b/src/extensions/score_source_code_linker/tests/test_source_code_link_integration.py
@@ -515,12 +515,14 @@ def test_source_link_integration_ok(
             expected_code_link = make_source_link(
                 example_source_link_text_all_ok[treq_id]
             )
-            actual_source_code_link = treq_info["source_code_link"]
+            actual_source_code_link = treq_info.get(
+                "source_code_link", "no source link"
+            )
             assert expected_code_link == actual_source_code_link, treq_id
 
             # verify testlinks
             expected_test_link = make_test_link(example_test_link_text_all_ok[treq_id])
-            actual_test_code_link = treq_info["testlink"]
+            actual_test_code_link = treq_info.get("testlink", "no test link")
             assert expected_test_link == actual_test_code_link, treq_id
     finally:
         app.cleanup()


dcalavrezo-qorix (approved):

AlexanderLanin:
- you have multiple authors here, do you want to keep it or squash and force push this change?
- could you shortly explain why it was unstable

AlexanderLanin:
Here are the ones that were blocked:

image

AlexanderLanin:
Ok, the rule is: `pull_request` requires manual approval. `pull_request_target` is executed immediately.

a-zw:
Why was it flaky? I'm not sure.

The asserts are weird. Apparently, the order in strings ("foo, bar" and "bar, foo") was not considered deterministic, so the string was converted into a set of characters for comparison which is rather impossible to debug. The order seems actually deterministic.

It isn't even clear to me _what_ we are testing here, so we might as well delete the test...

AlexanderLanin (approved):

PR #283, opened 2025-10-28
✨ Introduce needs-config-writer

ubmarco:
## 📌 Description
Introduction of
https://needs-config-writer.useblocks.com

This greatly helps with distributed configs for Sphinx-Needs and ubCode / ubc.
Also dynamic and static configuration can be merged.

Any user of the `docs()` function will get a complete ubproject.toml generated into their confdir.
This is a declarative duplicate of all the dynamically generated Sphinx-Needs configuration.
It makes it possible to use tools such as ubCode or ubc without any further setup.
Adding static shared configuration is also possible, see `shared.toml`.

This is also a nice mechanism to transpile the metadata.yaml to the Sphinx-Needs 6 schema validation.

Absolute bazel cache paths for needs.json or the path to the shared plantuml file are turned into relative paths that resolve once the documentation was built once.

Tested this manually for this repo (see ubproject.toml) and https://github.com/eclipse-score/score.
ubCode works code out-of-the box, even when needs.json dependencies are not there.

The generated files can be committed to git and maintained on every docs-as-code update that comes with config changes. This is the recommended way, as it enables users to use ubCode / ubc without ever running Bazel. It also works when switching between branches. The other option is to gitignore the file and generate it on the first run of `bazel run //:docs` (or one of the other `bazel run` docs build commands).

Screenshot of ubCode for the [score docs](https://github.com/eclipse-score/score/tree/main/docs) with the declarative configuration:
image

## 🚨 Impact Analysis

## ✅ Checklist

The existing tests use the new extension and cover it.
I have seen the CI failing during development due to this, and fixed all issues.

### DIFF
--- a/src/requirements.txt
+++ b/src/requirements.txt
@@ -35,7 +35,7 @@ babel==2.17.0 \
 basedpyright==1.29.2 \
     --hash=sha256:12c49186003b9f69a028615da883ef97035ea2119a9e3f93a00091b3a27088a6 \
     --hash=sha256:f389e2997de33d038c5065fd85bff351fbdc62fa6d6371c7b947fc3bce8d437d
-    # via -r /home/marco/.cache/bazel/_bazel_marco/3778152fef70679df5b348f45ca27594/external/score_tooling+/python_basics/requirements.txt
+    # via -r external/score_tooling+/python_basics/requirements.txt
 beautifulsoup4==4.13.4 \
     --hash=sha256:9bbbb14bfde9d79f38b8cd5f8c7c85f4b8f2523190ebed90e950a8dea4cb1c4b \
     --hash=sha256:dbb3c4e1ceae6aefebdaf2423247260cd062430a410e38c66f2baa50a8437195
@@ -452,7 +452,7 @@ iniconfig==2.1.0 \
     --hash=sha256:3abbd2e30b36733fee78f9c7f7308f2d0050e88f0087fd25c2645f63c773e1c7 \
     --hash=sha256:9deba5723312380e77435581c6bf4935c94cbfab9b1ed33ef8d238ea168eb760
     # via
-    #   -r /home/marco/.cache/bazel/_bazel_marco/3778152fef70679df5b348f45ca27594/external/score_tooling+/python_basics/requirements.txt
+    #   -r external/score_tooling+/python_basics/requirements.txt
     #   pytest
 jinja2==3.1.6 \
     --hash=sha256:0137fb05990d35f1275a587e9aee6d56da821fc83491a0fb838183be43f66d6d \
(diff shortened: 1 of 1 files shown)


docs/ubproject.toml
```diff
@@ -0,0 +1,1386 @@
...
+option = "uses"
+outgoing = "uses"
+
+[[needs.extra_links]]
+incoming = "violated_by"
+option = "violates"
+outgoing = "violates"
+
+[needs.flow_configs]
+score_config = "!include /home/marco/.cache/bazel/_bazel_marco/3778152fef70679df5b348f45ca27594/execroot/_main/bazel-out/k8-fastbuild/bin/docs.runfiles/_main/src/extensions/score_layout/assets/puml-theme-score.puml"
```
a-zw: Don't commit the ubproject.toml says the PR description because of the absolute paths.

ubmarco: I fixed this in the [new version](https://needs-config-writer.useblocks.com/changelog.html#release-0-2-4) of needs-config-writer. Paths are relative and stable now.
Even for the ones with a prefix like here `"!include <path>"`.

src/extensions/score_metamodel/external_needs.py
```diff
@@ -184,6 +184,9 @@ def connect_external_needs(app: Sphinx, config: Config):
             continue
 
         assert isinstance(app.config.needs_external_needs, list)  # pyright: ignore[reportUnknownMemberType]
+        # this sets the default value - required for the needs-config-writer
+        # setting 'needscfg_exclude_defaults = True' to see the diff
+        app.config.needs_external_needs = []
```
a-zw: The previous statement asserted it is a list and the next statement appends something. Why assign an empty list here? What black magic is going on here?

ubmarco: Yeah looks unconventional. But not black magic. I tried to express the rationale in the comment.
The Sphinx default value in app.config._options is set to the first value that the config receives.
I want this to be an empty list, so `needs-config-writer` sees the value is not at the default (which is an empty list then).
So *if* the value is set, it is actually put to the TOML file when setting https://needs-config-writer.useblocks.com/configuration.html#needscfg-exclude-defaults to `True`. We can look into this a bit more in future, but for now it works.

AlexanderLanin:
We don't have a proper solution for storing such config files, so while I don't like it, it would be consistent to just write that file into each repository for now.

But there is a scenario I really don't like here:
* Someone updates the docs-as-code dependency with some insignificant metamodel change.
* PR is green, all is good, it gets merged
* main branch is clean
* Now the next user comes along and builds docs. Suddenly they have a change in their local toml file.

AlexanderLanin (approved):

PR #286, opened 2025-10-31
Version 1.5.0 adds :html target

a-zw:
## 📌 Description

Offers an additional Bazel target for docs-as-code users to build inside the Bazel sandbox.

As a side effect, another commit on the main branch becomes part of this version increment. If we merge this first, then we can remove the git override in [this other PR](https://github.com/eclipse-score/reference_integration/pull/14/files?diff=unified&w=0#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR31).

## 🚨 Impact Analysis

Additional feature fully backwards compatible.

## ✅ Checklist

### DIFF
--- a/docs/how-to/commands.md
+++ b/docs/how-to/commands.md
@@ -8,6 +8,7 @@
 | `bazel run //:live_preview`                    | Creates a live_preview of the documentation viewable in a local server                            |
 | `bazel run //:live_preview_combo_experimental` | Creates a live_preview of the full documentation with all dependencies viewable in a local server |
 | `bazel run //:ide_support`                     | Sets up a Python venv for esbonio (Remember to restart VS Code!)                                  |
+| `bazel run //:html.serve`                      | Serves the generated HTML documentation on a local web server                                     |
 
 ## Internal targets (do not use directly)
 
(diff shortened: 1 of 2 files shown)


docs/how-to/commands.md
```diff
@@ -4,12 +4,14 @@
 
 | Target                      | What it does                                                           |
 | --------------------------- | ---------------------------------------------------------------------- |
-| `bazel run //:docs`         | Builds documentation                                                   |
+| `bazel build //:html`       | Builds documentation in Bazel sandbox                                  |
```
AlexanderLanin: We need a better name here, as having docs and html is confusing.

And I would not add a build target until we considered the other DRs. The effort and direction in the last months was to get away from bazel build. Maintaining build and run targets at the same time was enormous effort in the past.

We could add an "docs_experimental_sandbox" target, if we have any usecase for it. I don't want to open a lengthy discussion here in parallel to the DRs. But as it stands today, sandbox has only downsides for the sphinx build.

AlexanderLanin (changes requested):

PR #293, opened 2025-11-06
Remove id prefixes for mega-build (DR-004-Infra)

guejan:
## 📌 Description

- Remove all PROCESS_ prefixes from requirements and related files.
- Eliminate id_prefix from external needs configuration and code.
- Update tests and extensions to work without prefix logic.
- Make id_prefix handling optional in score_source_code_linker.

### Why:
S-CORE requires bi-directional traceability for compliance (see gd_req__req_traceability).
Previously, modules used prefixes for external needs, but this breaks integration when building a single, unified documentation site ("mega-build").

This change implements "Option S: Single documentation build" from decision record DR-004-Infra: Bi-directional traceability in docs (https://eclipse-score.github.io/score/main/design_decisions/DR-004-infra.html).

Option S was chosen because it enables quick integration and unified navigation, even though it results in longer build times and requires dropping prefixes.
Since our id schemas already provide namespacing, removing prefixes does not risk id clashes. This refactoring aligns with the decision to prioritize integration speed and documentation consistency for releases.

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/extensions/score_metamodel/external_needs.py
+++ b/src/extensions/score_metamodel/external_needs.py
@@ -191,7 +191,7 @@ def add_external_needs_json(e: ExternalNeedsSource, config: Config):
 
 
 def connect_external_needs(app: Sphinx, config: Config):
-    extend_needs_json_exporter(config, ["project_url", "project_prefix"])
+    extend_needs_json_exporter(config, ["project_url"])
 
     external_needs = get_external_needs_source(app.config.external_needs_source)
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,6 @@
 
 project = "Score Docs-as-Code"
 project_url = "https://eclipse-score.github.io/docs-as-code/"
-project_prefix = "DOCS_"
 version = "0.1"
 
 extensions = [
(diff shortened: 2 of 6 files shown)


src/extensions/score_metamodel/external_needs.py
```diff
@@ -183,18 +183,11 @@ def add_external_needs_json(e: ExternalNeedsSource, config: Config):
     assert isinstance(config.needs_external_needs, list)  # pyright: ignore[reportUnknownMemberType]
     config.needs_external_needs.append(  # pyright: ignore[reportUnknownMemberType]
         {
-            "id_prefix": needs_json_data["project_prefix"],
```
a-zw: The trick for a backwards-compatible change would probably be to assign an id_prefix here *unless some flag in conf.py is set to do otherwise*. Then every repo has to set DISABLE_EXTERNAL_NEEDS_PREFIXES=True. Once everybody has done so, we can actually remove this line. Then everybody can remove the DISABLE_EXTERNAL_NEEDS_PREFIXES from their conf.py again.

That sounds like too much work considering that "everybody" is effectively only the score Platform repo.

PR #346, opened 2026-01-15
test: add central S-Core PR checks workflow

AlexanderLanin:
This is just preparation. It does not yet replace any of the other workflows.

We can reduce the number of workflows once this central on_pr workflows handles some of them.


AlexanderLanin:
superseeded by #461

PR #394, opened 2026-02-06
Feat/generate ubproject and schema validation

arnoox:
## 📌 Description

- [x] Add sn_schemas.py which translates metamodel.yaml into a schemas.json file consumable by sphinx-needs. This enables schema-based validation at parse time: required fields, regex patterns on option values, mandatory link presence (minItems: 1), and mandatory link target type checks (validate.network).
- [x] Because ubCode (the VS Code extension for sphinx-needs) evaluates these schemas during editing, all metamodel violations now surface as IDE diagnostics directly in the editor, errors are caught early with lightweight, fast rendering, without waiting for a full Sphinx build.
- [x] Add unit tests (test_sn_schemas.py) and integration tests (test_sn_schemas_integration.py) that validate generated schemas against the real S-CORE metamodel using jsonschema_rs, the same engine sphinx-needs uses at runtime.
- [x] Add developer documentation (README.md) covering the two validation layers (schema vs. post-build Python checks) with a detailed coverage comparison table.

## 🚨 Impact Analysis

## ✅ Checklist


ubmarco (changes requested):
I'm missing a test for this. Ideally the generated schema.json is tested against the same SCORE metamodel tests. Then we can be sure the migration was successful.

arnoox:
Thanks for the review!

> I'm missing a test for this. Ideally the generated schema.json is tested against the same SCORE metamodel tests. Then we can be sure the migration was successful.

Yeah this was still work in progress. I introduced a refactored/more testable version of the sn_schemas transformation, including unit/integration tests.

arnoox:
The schema validation caught a new error in the score repo. We first need to fix it: https://github.com/eclipse-score/score/pull/2581

ubmarco (changes requested):
I tested this and it looks functionally working well. Here are some finding, a mix of AI, and my personal findings.

Found 2 issues:

1. The error message in `_classify_links` says only the **first** regex will be used when multiple regex patterns are supplied for the same link field, but the implementation unconditionally overwrites on each iteration so the **last** regex actually wins. The companion test even asserts the last value is kept (`# Last regex overwrites previous ones`), confirming the message is wrong.

   https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_metamodel/sn_schemas.py#L110-L120

   Test that confirms the actual (last-wins) behavior:

   https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_metamodel/tests/test_sn_schemas.py#L155-L163

   Fix: either change the message to say "the last one will be used", or `continue` after logging so the first one is actually retained (and update the test accordingly).

2. **`needs_metamodel_generated.toml` should not be written into `confdir`.** `_generate_needs_types_toml`, `_generate_needs_fields_toml` and `_generate_needs_links_toml` build TOML strings purely in memory, but the setup then writes the concatenation to `Path(app.confdir) / "needs_metamodel_generated.toml"` solely so the path can be appended to `needscfg_merge_toml_files`. `needs_config_writer` then merges the same content into `ubproject.toml`, so the on-disk file contains nothing the final `ubproject.toml` doesn't already have — and it lands inside the user's source tree (`docs/`), forcing a `.gitignore` entry. Running `bazel run //:docs_check` produces both files side-by-side with overlapping contents, which is confusing and serves no consumer.

   https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/__init__.py#L154-L167

   `.gitignore` entry added to hide the generated file (also covers the redundancy):

   https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/.gitignore

   README explicitly labels the file as an "Intermediate merge input for `needs_config_writer`":

   https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/README.md#L130-L137

   Fix options (in order of preference):
   - Write the intermediate to `app.outdir` (or a `tempfile.mkdtemp()` directory cleaned up via a `build-finished` handler) so it never appears under `confdir`. The `.gitignore` entry can then be removed.
   - Better still: extend `needs_config_writer` to accept an in-memory TOML fragment (string or dict) so the intermediate file is not needed at all. The current "deterministic path so no temp file leak" rationale (`__init__.py:157`) is solving the wrong problem — the leak only exists because `confdir` was chosen as the target.

---

## Lower-confidence findings (below the 80-point bar)

The 5 parallel review agents surfaced these additional candidates. Each was scored 0–100 by an independent agent; only items at ≥80 are reported as confirmed issues above. The rest are listed here for completeness — they are either latent (no current trigger), intentional, or pre-existing.

### 2. Optional-field regex alternation `^$|{pattern}` (score: 60)

In `get_pattern_schema`, optional fields wrap the metamodel pattern as `f"^$|{pattern}"`. JSON Schema's `pattern` keyword is defined as a substring search (ECMA-262 semantics, not full match), whereas the post-build Python checker uses `re.fullmatch`. For patterns that are not fully `^…$` anchored (e.g. `^https://github.com/.*` used in `source_code_link` / `testlink` fields in `metamodel.yaml`), the alternation `^$|^https://github.com/.*` may behave differently in `jsonschema-rs` (used by sphinx-needs / ubCode) than in the Python checker, allowing values one engine accepts and the other rejects.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_metamodel/sn_schemas.py#L300-L320

Why scored 60: The current metamodel mostly uses fully anchored patterns, so this is a real but latent inconsistency rather than a bug hitting today. Worth a TODO and a test for an unanchored pattern.

### 3. Missing TOML escaping in `_generate_needs_types_toml` (score: 60)

`_generate_needs_fields_toml` and `_generate_needs_links_toml` apply `.replace("\\", "\\\\").replace('"', '\\"')` before interpolating values into TOML strings. `_generate_needs_types_toml` does not — `directive`, `title`, and `prefix` are written raw. If any need-type title or prefix ever contains a backslash or double quote, the generated TOML will be syntactically invalid and downstream parsing will fail.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/__init__.py

Why scored 60: Real divergence from the sibling helpers, but no current metamodel value contains the problem characters, so the bug is latent. Easy to fix by mirroring the escape calls used in the field/link helpers.

### 4. Default ID pattern broadened to allow uppercase (score: 65)

The auto-generated fallback ID pattern in `yaml_parser.py` widened from `^{prefix}[0-9a-z_]+$` (lowercase only) to `^{prefix}[0-9a-zA-Z_]+$` (mixed case). This affects every need type that does not declare an explicit `id:` pattern.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_metamodel/yaml_parser.py#L130-L140

Why scored 65: The change appears to have ridden in on an earlier sphinx-needs 6.3.0 upgrade commit on the same branch (not introduced in this PR's primary diff hunks), and there is no commit message or comment explaining the policy shift. Confirm intentional; if so, document; if not, restore lowercase.

### 5. `needscfg_exclude_vars` may duplicate `needs_config_writer` defaults (score: 25)

In `score_sync_toml/__init__.py`, the setup code does `list(getattr(app.config, "needscfg_exclude_vars", None) or []) + ["needs_from_toml", "needs_from_toml_table", "needs_schema_definitions_from_json"]`. If `needs_config_writer` (loaded earlier in the extension order) already registers those three entries as the default value, the result will contain duplicates.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/__init__.py

Why scored 25: Could not confirm `needs_config_writer` actually registers these defaults; even if duplicates appear, exclusion-list duplication is typically benign. Worth a quick check during review.

### 6. Optional-link `validate.network` intentionally omitted (score: 5 — false positive)

`sn_schemas.py` deliberately skips emitting `validate.network` for optional links, with an in-source explanation: the post-build Python checker reports optional-link target-type violations at `treat_as_info=True`, while sphinx-needs schema validators only support a single `violation` severity per schema rule. Including optional links in the schema would escalate info-level findings to hard errors. A possible solution is to write optional links into dedicated rules, so they can be assigned a dedicated severity. The README's coverage table documents this trade-off.

Why scored 5: Documented intentional design decision with rationale and tests. Not an issue.

---

## Follow-up suggestions (not blocking this PR)

These are improvements to track in subsequent PRs — they are not regressions in #394, but the surrounding code makes them natural next steps.

### F1. Replace empty-string sentinels with nullable fields

`_generate_needs_fields_toml` always writes `default = ""` for any field that has no metamodel-supplied default. Throughout the codebase this empty string is then re-interpreted as "field not given" in later checks. Modelling absence with nullable fields (`null` / Optional) is the cleaner alternative — distinguishing a deliberately empty value from an unset one — and avoids the readability and type-safety problems that come with overloaded empty strings.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/__init__.py#L20-L43

### F2. Promote shared enum values to field-level schemas

Where a field carries the same enum across every need type that uses it, the per-type duplication can be hoisted into a single `[needs.fields.<name>]` schema entry, e.g.

```toml
[needs.fields.status]
schema.enum = ["draft", "review", "approved", "done"]
```

This would shrink the generated `schemas.json`, make ubCode autocompletion show enum values for the field everywhere it appears, and keep the metamodel DRY. `_generate_needs_fields_toml` currently emits only `default` — extending it with `schema.enum` (and possibly `schema.type`) when the same constraint holds across every type that uses the field is the natural follow-up.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/__init__.py#L20-L43

### F3. Emit `$schema` header in generated `ubproject.toml`

The generated `ubproject.toml` should start with

```toml
"$schema" = "https://ubcode.useblocks.com/ubproject.schema.json"
```

so that editors with TOML schema support give schema-aware completion and validation while editing the file. See [ubCode configuration docs](https://ubcode.useblocks.com/configuration/index.html). This would likely live as a top-of-file fragment in `shared.toml` or as a `needs_config_writer` option, depending on how the merge tool orders keys.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/shared.toml

### F4. Convert `needs_schema_debug_path` to a relative path

The generated `ubproject.toml` currently contains an absolute, machine-specific path:

```toml
schema_debug_path = "/home/marco/git/eclipse-score/docs-as-code/docs/schema_debug"
```

`needs_config_writer` already supports rewriting absolute paths via `needscfg_relative_path_fields`, and its own docstring explicitly names `needs_schema_debug_path` as the example field for this option:

> "List of config path patterns to convert to relative paths. Can be strings (e.g., `'needs_schema_debug_path'`) ..."
> — `needs_config_writer/main.py:94`

The fix is to add `"needs_schema_debug_path"` to the existing `needscfg_relative_path_fields.extend([...])` call. If sphinx-needs cannot consume a relative value for this field, that is an upstream bug to file against sphinx-needs / needs-config-writer; otherwise this is a one-line change.

https://github.com/eclipse-score/docs-as-code/blob/5f1263b1fccd9acde0d608e29e672f448e53979e/src/extensions/score_sync_toml/__init__.py#L168-L177

FScholPer (dismissed):

ubmarco (approved):
I am approving this PR, the noted review comments can be addressed in follow-up PRs. Functionally all good.

src/extensions/score_metamodel/README.md
```diff
@@ -0,0 +1,234 @@
+# score_metamodel
```
a-zw: I don't think it makes sense to create this README here which seems to be AI slop more or less duplicating info from our actual documentation.

src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -978,6 +978,12 @@ needs_extra_links:
   partially_verifies:
     incoming: partially_verified_by
     outgoing: partially_verifies
+
+  # Decision Records
+  affects:
+    incoming: affected by
+    outgoing: affects
```
a-zw, line 997: Why is the metamodel extended here? Was that a gap before? Why didn't we find that before then?

.gitignore
```diff
@@ -11,6 +11,8 @@ user.bazelrc
 # docs build artifacts
 /_build*
 docs/ubproject.toml
+docs/schemas.json
+docs/needs_metamodel_generated.toml
```
a-zw: Even more files getting generated into the workspace?

This clashes with my proposal here: https://github.com/eclipse-score/score/pull/2826

Technically, this is just a draft, so not really relevant for this PR though. Just want to mention it.

src/extensions/score_metamodel/tests/test_sn_schemas.py
```diff
@@ -0,0 +1,538 @@
...
+
+# =============================================================================
+# Tests for get_pattern_schema
+# =============================================================================
+
+
+class TestGetPatternSchema:
+    def test_returns_string_type_with_pattern(self) -> None:
+        result = get_pattern_schema("^[A-Z]+$")
+        assert result == {"type": "string", "pattern": "^[A-Z]+$"}
```
a-zw, line 40: We should trim down the worst of this AI generated stuff. Some tests here are not really useful.

AlexanderLanin:
`bazel build //:needs_json` fails with `WARNING: Failed to write schemas.json: [Errno 30] Read-only file system: '/home/alex/.cache/bazel/_bazel_alex/ec58ecca617edd0b2864120a40405cd7/sandbox/linux-sandbox/271/execroot/_main/bazel-out/k8-fastbuild/bin/_needs_json/_sources/docs/schemas.json'. Schema validation will be unavailable.`

Don't we test that in CI? 😨

src/extensions/score_metamodel/tests/test_sn_schemas_integration.py
```diff
@@ -0,0 +1,546 @@
...
+            "security": "YES",
+            "safety": "QM",
+            "status": "valid",
+            "satisfies": ["stkh_req__some_need"],
+        }
+
+    def test_valid_need_passes(
+        self, schemas_by_type: dict[str, dict[str, Any]]
+    ) -> None:
+        assert_schema_valid(self._make_valid(), schemas_by_type["feat_req"])
```
a-zw, line 295: Same as in the other test file: Some tests are pointless and should be removed.

src/extensions/score_metamodel/__init__.py
```diff
@@ -248,6 +250,22 @@ def setup(app: Sphinx) -> dict[str, str | bool]:
     config_setdefault(app.config, "needs_reproducible_json", True)
     config_setdefault(app.config, "needs_json_remove_defaults", True)
 
+    # Generate schemas.json from the metamodel and register it with sphinx-needs.
+    # This enables sphinx-needs 6 schema validation: required fields, regex
+    # patterns on option values, and (eventually) link target type checks.
+    # Use config-inited event to defer file writing until config is ready,
+    # with error handling to prevent boot crashes on file write failures.
```
a-zw: Shouldn't his be a docstring? Comment before method makes no sense.

a-zw (changes requested):

src/extensions/score_metamodel/metamodel.yaml
AlexanderLanin: I was just reviewing that. It was a gap. Why we didnt find that ourselves is a very valid question.

.gitignore
arnoox: - `docs/needs_metamodel_generated.toml` can definitely be moved to some build dir. I'll fix as part of this PR.
- `docs/schemas.json` is maybe required in the workspace. I can check if it is possible for `ubproject.toml` to reference to something from within the build dir.

src/extensions/score_metamodel/README.md
arnoox: fair point. I will remove this.

src/extensions/score_metamodel/tests/test_sn_schemas.py
arnoox: Agreed, the tests for the trivial dict building helpers add no real value. I will keep only the tests that cover actual branching logic.

src/extensions/score_metamodel/__init__.py
arnoox: Fixed.

AlexanderLanin:
I'm not so sure about the architecture here. score_metamodel writes the json, and the toml is written by a separate extension. However the toml has a hard dependency on having a json file.

Don't the toml and the json also overlap in their content?

AlexanderLanin:
Can we figure out some common naming scheme for the generated files? With this change we'll need to touch gitignore for all users, so some wildcard pattern would be nice.

src/extensions/score_metamodel/metamodel.yaml
arnoox: In fact, this comes from here: https://github.com/eclipse-score/docs-as-code/pull/394/changes/BASE..5f1263b1fccd9acde0d608e29e672f448e53979e#diff-0ac6046ac75bc866580d21188af43573b945253e760280edbf05da368090304bR874

The introduced transformation script was not happy about not finding the definition of the optional link.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
...
+        link_values = [v.strip() for v in value.split(",")]
+        for link_value in link_values:
+            if link_value.startswith("^"):
+                if field in regexes:
+                    LOGGER.error(
+                        f"Multiple regex patterns for {label} link field "
+                        f"'{field}' in need type '{type_name}'. "
+                        "Only the first one will be used in the schema."
+                    )
+                regexes[field] = link_value
```
AlexanderLanin, line 132: AI review note: Bug in log message — says "Only the first one will be used" but the code overwrites with `regexes[field] = link_value`, so the **last** regex wins. The unit test at `test_sn_schemas.py:602` confirms this (`regexes == {"field": "^regex2$"}`). Should read "Only the last one will be used".

src/extensions/score_metamodel/__init__.py
```diff
@@ -248,6 +250,24 @@ def setup(app: Sphinx) -> dict[str, str | bool]:
...
+        Use config-inited event to defer file writing until config is ready,
+        with error handling to prevent boot crashes on file write failures.
+        """
+        try:
+            write_sn_schemas(app, metamodel)
+        except Exception as e:
+            logger.warning(
+                f"Failed to write schemas.json: {e}. "
+                "Schema validation will be unavailable."
+            )
```
AlexanderLanin, line 335: AI review note: `except Exception` here silently degrades on programming errors or invalid metamodel data, not just file-write failures. Consider catching `OSError` specifically (the actual failure mode: disk/permission issues) and letting other exceptions propagate so bugs stay visible during development.

src/extensions/score_sync_toml/__init__.py
```diff
@@ -17,35 +17,155 @@
...
+    Returns only the fields that ubCode requires to identify need directives
+    (``directive``, ``title``, ``prefix``, and optionally ``color``/``style``).
+
+    Must be called *after* ``score_metamodel.setup()`` has run (i.e. after
+    ``app.config.needs_types`` has been extended with the metamodel types).
+    """
+    lines: list[str] = []
+    for nt in app.config.needs_types:
+        lines.append("[[needs.types]]\n")
+        lines.append(f'directive = "{nt["directive"]}"\n')
```
AlexanderLanin: AI review note: Field names, link names, and directive names are interpolated directly into TOML key syntax (e.g. `f"[needs.fields.{field_name}]"`). If a name ever contains a dot, space, or quote, the output is malformed TOML. Currently safe given metamodel content, but fragile.

Suggestion: add an assertion like `assert re.match(r"^[a-zA-Z0-9_]+$", name)` at the top of each generator, so a bad name fails loudly instead of producing silently broken TOML.

src/extensions/score_metamodel/metamodel.yaml
AlexanderLanin, line 997: AI review note: This `affects` link addition is a metamodel extension unrelated to schema generation. The author explained the schema script surfaced a pre-existing gap, which is fair context. For clean traceability, consider splitting this into its own commit so the metamodel change and the tooling change are independently revertable.

AlexanderLanin (commented):
AI-assisted review (Claude). Four findings: one bug, two robustness concerns, one scope note.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
+# *******************************************************************************
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
```
MaximilianSoerenPollak, line 2: Should it not be 2026 copyright? This file is completely new.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
...
+from sphinx.application import Sphinx
+from sphinx.config import Config
+from sphinx_needs import logging
+
+from src.extensions.score_metamodel.metamodel_types import ScoreNeedType
+from src.extensions.score_metamodel.yaml_parser import MetaModelData
+
+# Fields whose values are lists in sphinx-needs (e.g. tags: ["safety", "security"]).
+# These need an "array of strings" JSON schema instead of a plain "string" schema.
+SN_ARRAY_FIELDS = {
```
MaximilianSoerenPollak, line 38: All links to other needs are lists as well no? 
Like `satisfied_by` etc. 
So all of these should be added here or am I missing something?

src/extensions/score_metamodel/sn_schemas.py
MaximilianSoerenPollak: Ye. This is an issue. 
If you only want to use the first one you have to break after or you use the last one and redo the comment.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
...
+    for field, pattern in mandatory_fields.items():
+        if field in IGNORE_FIELDS:
+            continue
+        required.append(field)
+        properties[field] = get_field_pattern_schema(field, pattern, is_optional=False)
+
+    # Optional fields: if present, must match the regex pattern
+    # Allow empty strings to align with Python checker behavior
+    # Skip fields already handled as mandatory to avoid pattern/required conflicts
+    # when a field appears in both mandatory_options and optional_options
```
MaximilianSoerenPollak, line 168: A field should not be allowed in both mandatory & optional options. That makes no sense to have that.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
...
+    # (e.g. global base options merged into every type's optional_options).
+    for field, pattern in optional_fields.items():
+        if field in IGNORE_FIELDS:
+            continue
+        if field in mandatory_fields:
+            continue
```
MaximilianSoerenPollak, line 174: Should never happen. 

In my opinion this should be a full on exit. This does not exist in our world view and therefore should crash to ensure the person fixes it.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
...
+        return None
+
+    type_name = need_type["directive"]
+
+    # Classify link values as regex patterns vs. target type names.
+    # Note: links are still plain strings at this point (before postprocess_need_links).
+    mandatory_links_regexes, mandatory_links_targets = _classify_links(
+        mandatory_links, type_name, mandatory=True
+    )
+    optional_links_regexes, _ = _classify_links(
```
MaximilianSoerenPollak, line 230: So optional link targets are thrown away here. Is there a reason?

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -0,0 +1,328 @@
...
+    # 1. Mixed regex+plain fields (e.g.
+    #    "complies: std_wp, ^std_req__aspice_40__iic.*$"):
+    #    The items schema would incorrectly require ALL linked needs to match
+    #    the plain type, while some legitimately match the regex instead.
+    #
+    # 2. Optional links: The Python validate_links() in check_options.py treats
+    #    optional link type violations as informational (treat_as_info=True),
+    #    but schemas use a single severity ("violation") per need type.
+    #    Including optional links would escalate info-level issues to errors.
+    #    Optional link types are validated by the Python check instead.
```
MaximilianSoerenPollak, line 267: This will be changed shortly. 
Optional Links are now getting bit by bit also checked & will make a build error in the end. 

So this should then work the same here as well.

src/extensions/score_sync_toml/__init__.py
```diff
@@ -17,35 +17,155 @@
...
+        "needs_schema_definitions",
+        # needs_render_context contains Python callable objects (draw_full_interface,
+        # draw_full_module, etc.) that cannot be serialized to TOML.
+        # needs_config_writer emits unsupported_type warnings for these and skips them.
+        "needs_render_context",
+    ]
+
+    # Merge the static TOML file into the generated configuration.
     app.config.needscfg_merge_toml_files.append(
         str(Path(__file__).parent / "shared.toml")
```
MaximilianSoerenPollak, line 170: Idk if this will work correctly inside bazel with the parent of __file__.

MaximilianSoerenPollak (commented):
Just some comments I saw. 

Probably need further review when I understand more / have more time.

src/extensions/score_metamodel/tests/test_sn_schemas.py
```diff
@@ -348,8 +348,12 @@ def test_optional_links_with_regex(self) -> None:
 
 class TestWriteSnSchemas:
     def test_writes_json_file(self, tmp_path: Path) -> None:
+        outdir = tmp_path / "output"
+        confdir = tmp_path / "conf"
+        confdir.mkdir()
         app = MagicMock()
```
MaximilianSoerenPollak, line 354: You are using mocks here. 

Are you 100% sure that this is **actually** testing the implementation and the universe in where the application runs, not just the mock / mock universe?

src/extensions/score_metamodel/tests/test_sn_schemas.py
```diff
@@ -348,8 +348,12 @@ def test_optional_links_with_regex(self) -> None:
...
-        app.confdir = str(tmp_path)
+        app.outdir = str(outdir)
+        app.confdir = str(confdir)
```
MaximilianSoerenPollak, line 356: Nit (not merge blockers): 

Kinda useless assignment here you could just do: 

`app.outdir = str(tmp_path / "output")` etc.

src/extensions/score_metamodel/sn_schemas.py
```diff
@@ -65,13 +69,23 @@ def write_sn_schemas(app: Sphinx, metamodel: MetaModelData) -> None:
...
+    # Tell sphinx-needs to load the schema using the absolute path (outdir).
+    config.needs_schema_definitions_from_json = str(schemas_output_path)
+
+    # Best-effort copy to confdir so ubCode can find it at edit time.
+    try:
+        confdir_path = Path(app.confdir) / "schemas.json"
+        with open(confdir_path, "w", encoding="utf-8") as f:
+            json.dump(schema_definitions, f, indent=2, ensure_ascii=False)
+    except OSError:
+        pass  # Read-only filesystem (e.g. Bazel sandbox) — ubCode copy skipped.
```
MaximilianSoerenPollak, line 88: Is this silent fail for OSError wanted? 
Might be good to have a debug log here so you have some logs if wanted to see this branch is hit.

PR #431, opened 2026-03-03
Any folder in docs

a-zw:
## 📌 Description

A proposal for a mechanism how to include other directories of the repo into documentation.

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/extensions/docs/any_folder.rst
+++ b/src/extensions/docs/any_folder.rst
@@ -1,3 +1,17 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2026 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
 Any Folder
 ==========
 
--- a/src/extensions/docs/index.rst
+++ b/src/extensions/docs/index.rst
@@ -1,15 +1,16 @@
-..  # *******************************************************************************
-    # Copyright (c) 2025 Contributors to the Eclipse Foundation
-    #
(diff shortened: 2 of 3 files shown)


a-zw:
> * static .md file that serves as template and is commited. CICD during test execution puts some some data in in
> * detailed coverage reports which are only temp artifacts in the workflow. I would like to add them to `_build` directory and create links in static .rst pages to them.
> 
> First option can also get rid of the template and use only temp artifact if possible.

That sounds like there is already some scripting around both use cases anyways. Wouldn't it be easier to just extend these script a little and copy/symlink the stuff into `docs/`?

src/extensions/score_any_folder/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+    """Return ``(resolved_source, link_path)`` pairs from the mapping."""
+    confdir = Path(app.confdir)
+    pairs = []
+    for source_rel, target_rel in app.config.score_any_folder_mapping.items():
+        if Path(source_rel).is_absolute():
+            logger.error(
+                "score_any_folder: source path must be relative, got: %r; skipping",
+                source_rel,
+            )
+            continue
```
MaximilianSoerenPollak: Should this truly be an error then? 
The build will be broken if thati s the case, that is wanted here I guess?

src/extensions/score_any_folder/__init__.py
```diff
@@ -0,0 +1,118 @@
...
+        source = (confdir / source_rel).resolve()
+        link = confdir / target_rel
+        pairs.append((source, link))
+    return pairs
+
+
+def _create_symlinks(app: Sphinx) -> None:
+    created_links: set[Path] = set()
+
+    for source, link in _symlink_pairs(app):
```
MaximilianSoerenPollak: Does this work? 
You get a `list[tuple[Path,Path]]` back. But that list could have one or many tuples inside. 
Will this actually unwrap each tuple inside it? 
Or is the `symlink_pairs` function guranteed to only return 1 tuple?

MaximilianSoerenPollak (approved):
All comments have been addressed. 

Thanks for the work

PR #449, opened 2026-03-24
Allow to configure a different metamodel

### LINKED ISSUE #415: Extensible metamodel
When reusing docs-as-code for internal/proprietary purposes, I need to extend the metamodel and its checks.

While it is possible to some degree with some Python hackery, there should be a clean and documented way to do this.

The current metamodel extension also applies checks to types not defined for the S-CORE metamodel. For example, the id schema is enforced for any type.

a-zw:
Fixes https://github.com/eclipse-score/docs-as-code/issues/415

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -125,7 +125,7 @@ def _missing_requirements(deps):
         fail(msg)
     fail("This case should be unreachable?!")
 
-def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good = None):
+def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good = None, metamodel = None):
     """Creates all targets related to documentation.
 
     By using this function, you'll get any and all updates for documentation targets in one place.
@@ -135,12 +135,23 @@ def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good =
       data: Additional data files to include in the documentation build.
       deps: Additional dependencies for the documentation build.
       scan_code: List of code targets to scan for source code links.
+      known_good: Optional label to a "known good" JSON file for source links.
+      metamodel: Optional label to a metamodel.yaml file. When set, the extension loads this
+                 file instead of the default metamodel shipped with score_metamodel.
     """
 
     call_path = native.package_name()
 
     if call_path != "":
         fail("docs() must be called from the root package. Current package: " + call_path)
+
+    metamodel_data = []
+    metamodel_env = {}
+    metamodel_opts = []
+    if metamodel != None:
+        metamodel_data = [metamodel]
+        metamodel_env = {"SCORE_METAMODEL_YAML": "$(location " + str(metamodel) + ")"}
+        metamodel_opts = ["--define=score_metamodel_yaml=$(location " + str(metamodel) + ")"]
 
     module_deps = deps
     deps = deps + _missing_requirements(deps)
@@ -152,7 +163,7 @@ def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good =
     sphinx_build_binary(
         name = "sphinx_build",
         visibility = ["//visibility:private"],
-        data = data,
+        data = data + metamodel_data,
         deps = deps,
     )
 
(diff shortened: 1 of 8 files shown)


a-zw:
Chat with Max: Looks like the right direction.

Wait until after 0.7 release before merging this.

src/extensions/score_metamodel/tests/test_metamodel__init__.py
```diff
@@ -21,9 +25,15 @@
...
     parse_checks_filter,
+    setup,
 )
 from src.extensions.score_metamodel.tests import need
 
+# The module where setup() looks up load_metamodel_data at runtime.
+# Using __globals__ avoids breakage when the same file is imported under
+# two different names (e.g. "score_metamodel" and "src.extensions.score_metamodel").
+_setup_module = sys.modules[setup.__module__]
+
```
MaximilianSoerenPollak: I don''t quiet understand this. 
Where is it that you would import this in 2 different ways?

src/extensions/score_metamodel/tests/test_metamodel__init__.py
```diff
@@ -176,3 +186,58 @@ def test_combined_core_links_and_extras(self):
         assert n.get("output", []) == ["output_wp"]
         # Extra fields
         assert n.get("custom_attr") == "custom_value"
+
```
MaximilianSoerenPollak: Not a fan of mocks as it truly is an antipattern. 

Are you sure you are testing the actual implementation /universe here, and not just the mock implementation?

If that is the case I guess it's okay, though still not ideal.

src/extensions/score_metamodel/yaml_parser.py
```diff
@@ -182,11 +182,16 @@ def _collect_all_custom_options(
...
     """
     Load metamodel.yaml and prepare data fields as needed for sphinx-needs.
+
+    Args:
+        yaml_path: Path to the metamodel YAML file. When None, the default
+                   metamodel shipped with this extension is used.
     """
-    yaml_path = Path(__file__).resolve().parent / "metamodel.yaml"
+    if yaml_path is None:
+        yaml_path = Path(__file__).resolve().parent / "metamodel.yaml"
```
MaximilianSoerenPollak, line 194: Since we have the metamodel schema as well , could we directly after reading check if it confirms to the schema and if not early exit? 
What do you think of this idea?

docs.bzl
```diff
@@ -212,19 +223,29 @@ def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good =
         name = "docs",
         tags = ["cli_help=Build documentation:\nbazel run //:docs"],
         srcs = ["@score_docs_as_code//src:incremental.py"],
-        data = docs_data,
+        data = data + [":sourcelinks_json"] + metamodel_data,
```
MaximilianSoerenPollak: Wondering why the docs_env was removed & split up again? Or is that just an accident

MaximilianSoerenPollak (commented):
Some questions, but overall looks good and seems like a good feature to add.

src/extensions/score_metamodel/yaml_parser.py
a-zw: You want to enforce that it is always a metamodel extension and never modifies it?

I would consider it a valid use case to relax or constrain some things. For example, allow short ids?

docs.bzl
a-zw: Looks like an accident to me. We should probably revise the docs_env and docs_sources_env and stuff altogether. ToDo.

src/extensions/score_metamodel/yaml_parser.py
a-zw: We discussed the design decision "extension vs replacement" and want to go for replacement.

src/extensions/score_metamodel/tests/test_metamodel__init__.py
a-zw: Removed them. These tests were not useful enough.

docs.bzl
a-zw: Updated and simplified again.

AlexanderLanin (approved):

PR #451, opened 2026-03-25
Update to Sphinx-Needs 7.0.0

### LINKED ISSUE #442: Upgrade Sphinx Needs version
### What

We are currently on version 6.3. 

Version 7 has come out recently and some things that we want like versioning of needs will be available in 8.x

So we should upgrade in the near future. 

I have a inkling that this will not be as easy as I hope

- [x] Upgrade to Sphinx-Needs 7.x
- [ ] Upgrade to Sphinx-Needs 8.x

### Acceptance Criteria (DoD)

When sphinx-needs version is the currently latest.

### How

Upgrade Sphinx-Needs to newest available version

a-zw:
- Migrate from deprecated needs_extra_options/needs_extra_links to needs_fields/needs_links
- replace needs_global_options with needs_default_layout
- remove obsolete filterwarnings

Fixes https://github.com/eclipse-score/docs-as-code/issues/442


src/requirements.txt
```diff
@@ -806,6 +806,20 @@ mdurl==0.1.2 \
     --hash=sha256:84008a41e51615a49fc9966191ff91509e3c40b939176e643fd50a5c2196b8f8 \
```
MaximilianSoerenPollak: How is sphinx-needs itself not being updated in the requirements.txt? Seems very weird.

a-zw: Correct. The square brackets are for optional extra features.

From Sphinx-Needs:

> If you wish to also use the plotting features of sphinx-needs (see needbar and needpie), you need to also install matplotlib, which is available via the plotting extra:
>
> pip install sphinx-needs[plotting]

MaximilianSoerenPollak (approved):
For me this looks good.

Issue #459, opened 2026-03-31
Enable versioning of requirements


MaximilianSoerenPollak:
### What

Since sphinx-needs 8 enabels versioning of requirements now (via filterstrings in links). 
And we just upgraded to it, we should enable this and see how we can do this in our metamodel the best way.

### Acceptance Criteria (DoD)

Verisoning works with our metamodel, is tested & has documentation

PR #462, opened 2026-04-01
Enable versioning of needs

MaximilianSoerenPollak:
## 📌 Description
Adds the pre-requisites of versioning and a non-fatal check to see if needs have version attribute. 
Closes: #459 

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs/internals/requirements/requirements.rst
+++ b/docs/internals/requirements/requirements.rst
@@ -1148,18 +1148,6 @@ Overview of Tool to Process Requirements
 .. ------------------------------------------------------------------------
 ..
 
-Grouped Requirements
-####################
-
-.. tool_req:: Metamodel
-  :id: tool_req__docs_metamodel
-  :tags: metamodel
-  :version: 1
-  :implemented: YES
-
-  Docs-as-Code shall provide a metamodel for definining config in a  `metamodel.yaml` in the source code repository.
-
-  .. note:: "satisfied by" is something like "used by" or "required by".
 
 
 .. needextend:: c.this_doc() and type == 'tool_req'
@@ -1168,7 +1156,3 @@ Grouped Requirements
 
 .. needextend:: c.this_doc() and type == 'tool_req' and not status
   :status: valid
-
-.. needextend:: "metamodel.yaml" in source_code_link
-  :+satisfies: tool_req__docs_metamodel
-  :+tags: config
--- a/docs/internals/extensions/source_code_linker.md
+++ b/docs/internals/extensions/source_code_linker.md
@@ -12,6 +12,7 @@ The feature is split into **two layers**:
 
 1. **Bazel pre-processing (before Sphinx runs)**
    Generates and aggregates JSON caches containing the raw `source_code_link` findings (and, if available, repository metadata like `repo_name/hash/url`).
+   Note that "hash" might also be a "version" tag.
 
 2. **Sphinx extension (during the Sphinx build)**
    Reads the aggregated JSON, validates and adapts it, and finally injects the links into Sphinx needs in the final layout (**RepoSourceLink**).
@@ -36,6 +37,7 @@ one JSON cache per repository.
 It also adds metadata to each needlink that is needed in further steps.
 
 Example of requirement tags:
+
 ```python
 # Choose one or the other, both mentioned here to avoid detection
 # req-Id/req-traceability: <NEED_ID>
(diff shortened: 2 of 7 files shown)


src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -105,6 +105,37 @@ def _validate(attributes_to_allowed_values: dict[str, str], mandatory: bool):
...
+    """
+    This is a temporary check to allow for non fatal warnings if a version attribute
+    is missing in any need.
+    In future releases the version will be mandatory and this check
+    can be deleted as then it is covered by the normal validate_options function
+    Checks if version attribute is present in the need. Will emit a non fatal warning
+    """
+    need_version = need.get("version")
+    if not need_version:
+        log.warning_for_need(
```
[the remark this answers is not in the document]
MaximilianSoerenPollak: I think this might be the better approach tbh. To have it work 'silently' as you said. And then in the next version add these warnings slowly in. 
Will change this accordingly.

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -105,6 +105,37 @@ def _validate(attributes_to_allowed_values: dict[str, str], mandatory: bool):
...
+            + "are required to have the version attribute in future releases. "
+            + "It should be a whole number like: 'version: 1'.",
+            is_new_check=True,
+        )
+    elif not need_version.isdigit():
+        # As the version is already in the need here, this can be a 'error log'.
+        log.warning_for_need(
+            need,
+            msg="Version is required to be a whole number e.g '1, 10, 12'",
+            is_new_check=True
```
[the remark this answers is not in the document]
MaximilianSoerenPollak: Ah, ye that indeed should be the case. 

Changed in newest version.

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -105,6 +105,45 @@ def _validate(attributes_to_allowed_values: dict[str, str], mandatory: bool):
     _validate(need_type["optional_options"], False)
 
 
+#              ╭──────────────────────────────────────────────────────────╮
+#              │ Checks will be deactivated for now to give silent grace  │
+#              │                         period.                          │
+#              │    Afterwards they will be activated as 'non fatal'.     │
```
a-zw, line 111: Why not immediately as non-fatal?

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -105,6 +105,45 @@ def _validate(attributes_to_allowed_values: dict[str, str], mandatory: bool):
...
+#     if not need_version:
+#         log.warning_for_need(
+#             need,
+#             msg="is missing attribute `version`. All needs "
+#             + "are required to have the version attribute in future releases. "
+#             + "It should be a whole number like: 'version: 1'.",
+#             is_new_check=True,
+#         )
+#     elif not need_version.isdigit():
+#         # As the version is already in the need here, this can be a 'error log'.
```
a-zw, line 140: Wouldn't it make sense to have this one immediately as fatal?

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -105,6 +105,45 @@ def _validate(attributes_to_allowed_values: dict[str, str], mandatory: bool):
...
+#     need_version = need.get("version")
+#     if not need_version:
+#         log.warning_for_need(
+#             need,
+#             msg="is missing attribute `version`. All needs "
+#             + "are required to have the version attribute in future releases. "
+#             + "It should be a whole number like: 'version: 1'.",
+#             is_new_check=True,
+#         )
+#     elif not need_version.isdigit():
```
a-zw, line 139: Do we allow "0"?

To be precise we should adapt tool_req__docs_common_attr_version to say "natural numbers excluding 0".

docs/internals/requirements/requirements.rst
a-zw, line 250: Excluding 0?

src/extensions/score_metamodel/checks/check_options.py
AlexanderLanin: See comment above. My suggestion was to support early adopters first, until you are 120% sure that it works. Before telling everyone to add versions.

docs/internals/requirements/requirements.rst
AlexanderLanin: Why exactly without 0? Personally I like v0 as indication of an very early stage.

src/extensions/score_metamodel/checks/check_options.py
MaximilianSoerenPollak: Hmm I think 0 should be dissallowed, but this should be determined in process right? 
For now all of the checks are commented out so this still has time to be determined once we include them again.

docs/internals/requirements/requirements.rst
MaximilianSoerenPollak: I think for now it allows anything, and then the process community has to clarify if 0 is allowed or not? 
And then we can implement it.

a-zw: For me 0 would be equivalent to status=draft and thus it is pointless.

PR #467, opened 2026-04-02
Add sphing toolchain

ramceb:
## 📌 Description
This PR adds a bazel toolchain for sphinx builds, 
which allows to inject the binary which is used 
for build html documentation

## 🚨 Impact Analysis

- [X] This change does not violate any tool requirements and is covered by existing tool requirements
- [X] This change does not violate any design decisions

## ✅ Checklist

- [X] Added/updated documentation for new or changed features
- [X] Followed project coding standards and guidelines

### DIFF
--- a/bazel/toolchains/BUILD
+++ b/bazel/toolchains/BUILD
@@ -1,0 +1,45 @@
+load(
+    "@score_tooling//bazel/rules/rules_score:sphinx_toolchain.bzl",
+    "sphinx_toolchain",
+)
+# ============================================================================
+# Toolchain for the score build
+# ============================================================================
+py_binary(
+    name = "score_build",
+    srcs = ["@score_tooling//bazel/rules/rules_score:src/sphinx_wrapper.py"],
+    data = [],
+    env = {
+        "SOURCE_DIRECTORY": "",
+        "DATA": "",
+        "ACTION": "check",
+    },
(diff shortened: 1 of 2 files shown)


AlexanderLanin:
Having something overwritable/injectable seems ok. However this adds yet another dependency. I would rather remove the existing dependencies. But of course I dont know how to do that 😆

ramceb:
> Having something overwritable/injectable seems ok. However this adds yet another dependency. I would rather remove the existing dependencies. But of course I dont know how to do that 😆
The problem was the circular dependency between tooling and docs-as-code. We removed the dependency from tooling to docs-as-code in https://github.com/eclipse-score/tooling/pull/116. 
The already existing dependency from docs-as-code to tooling we need to keep. So the dependency chain looks as follows
```
module which wants to generate documentation
            ↓
       docs-as-code
            ↓
          tooling
```
docs-as-code defines the score_toolchain incl. all of the metamodel for generating the documentation

bazel/toolchains/BUILD
```diff
@@ -0,0 +1,45 @@
...
+)
+
+sphinx_toolchain(
+    name = "score_sphinx_toolchain",
+    conf_template = "@score_tooling//bazel/rules/rules_score:templates/conf.template.py",
+    sphinx = ":score_build",
+)
+
+toolchain(
+    name = "score_toolchain",
```
a-zw, line 33: This toolchain/file is not used so far? Why do we need it?

PR #471, opened 2026-04-07
known_good.json maybe use "version" instead of "hash"

a-zw:
## 📌 Description

The known_good.json file may use "version" instead of "hash" and that must be supported.

### DIFF
--- a/docs/internals/extensions/source_code_linker.md
+++ b/docs/internals/extensions/source_code_linker.md
@@ -12,6 +12,7 @@ The feature is split into **two layers**:
 
 1. **Bazel pre-processing (before Sphinx runs)**
    Generates and aggregates JSON caches containing the raw `source_code_link` findings (and, if available, repository metadata like `repo_name/hash/url`).
+   Note that "hash" might also be a "version" tag.
 
 2. **Sphinx extension (during the Sphinx build)**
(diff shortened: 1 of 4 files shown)


docs/internals/extensions/source_code_linker.md
```diff
@@ -12,6 +12,7 @@ The feature is split into **two layers**:
```
MaximilianSoerenPollak, line 12: I'm guessing these changes come from a formatter of some sort? 
We probably should agree to a formatter to use for markdown. 
As I see it most of these changes do make sense and seem sensible and good.

MaximilianSoerenPollak, line 12: I'm guessing these changes come from a formatter of some sort?
We probably should agree to a formatter to use for markdown.
As I see it most of these changes do make sense and seem sensible and good.

MaximilianSoerenPollak (approved):
Looks good. 

Seems like we only needed a small changed indeed.:

only one comment though that has not much to do with the PR itself.

docs/internals/extensions/source_code_linker.md
a-zw: It was manual actually. VSCode annoyed me with lots of warnings, so i fixed the easy ones. 🤷‍♂️

I agree that we should certainly not use different styles. As an alternative, we could just convert the md files to rst.

Anyways, not in this PR, I think.

PR #474, opened 2026-04-09
chore: drop multitool & cleanup

AlexanderLanin:
A while ago we decided to not use bazel multitool, and docs-as-code is the only repo thats using multitool. So lets remove that.

* as discussed shellcheck is dropped, as no shell scripts are in the repo
* dependency to multitool is removed
* dependency to tooling is smaller (one step at a time)


MODULE.bazel
```diff
@@ -12,11 +12,7 @@
 # *******************************************************************************
 
 # Versioning is handled by GitHub Releases.
-# Here we always keep "0.0.0" to indicate that.
-module(
-    name = "score_docs_as_code",
-    version = "0.0.0",
-)
+module(name = "score_docs_as_code")
```
MaximilianSoerenPollak, line 15: no version?

src/tests/test_consumer.py
```diff
@@ -264,7 +264,7 @@ def parse_bazel_output(BR: BuildOutput, pytestconfig: Config) -> BuildOutput:
     if pytestconfig.get_verbosity() >= 2 and os.getenv("CI"):
         console.print("[DEBUG] Raw warnings in CI:")
         for i, warning in enumerate(split_warnings):
-            console.print(f"[DEBUG] Warning {i}: {repr(warning)}")
+            console.print(f"[DEBUG] Warning {i}: {warning!r}")
```
MaximilianSoerenPollak: This `!r` stuff seems weird to me and less readable. 
It is apprently a 'backwards compatibitliy thing to have it still work with 'str.format'.

AlexanderLanin: excluded from this PR

MODULE.bazel
AlexanderLanin: Apparently thats best practice now. It's supported by bazel_registry since yesterday. You reviewed the change :p

MaximilianSoerenPollak: Ah I thought that was only compatibility_level that was outdated. But ye I remember.

pyproject.toml
```diff
@@ -1,27 +1,76 @@
...
 ]
+
+
+[tool.ruff.lint.isort]
+combine-as-imports = true
+
+[tool.ruff.lint.per-file-ignores]
+"tests/**" = [
+  "S101",  # assert used
+]
```
MaximilianSoerenPollak, line 71: Since we have in many places asserts used to 'assert' the worldview so to speak. I wonder if this needs to be removed completely? 
though not sure.

AlexanderLanin: We'll see once that warning starts to show up everywhere :D

MaximilianSoerenPollak (approved):
Makes things for sure simpler in the long run. 
Little by little removing dependencies to other modules. 

Thanks 💯

PR #476, opened 2026-04-09
chore: use py_venv directly

AlexanderLanin:
Reducing dependency to tooling bit by bit


MaximilianSoerenPollak (approved):
Ye this makes a ton of sense after re-evaluation.

PR #477, opened 2026-04-09
chore: use copyright checker directly without bazel

AlexanderLanin:
Reducing dependency to tooling bit by bit

Funny enough the new approach has added a copyright to pre-commit yaml 😆


MaximilianSoerenPollak (dismissed):
Making it simpler still. 💯

MaximilianSoerenPollak (approved):
Looks good to me. 

Kinda wondering how so many places didn't have the copyright. Good that we added it now.

PR #481, opened 2026-04-10
chore: drop dependency to score_tooling

AlexanderLanin:
last PR to drop tooling dependency

side-effect: we loose cli-helper

PR #482, opened 2026-04-10
fix: support docs at root level

AlexanderLanin:
docs at root level are impossible without this change

note: merge https://github.com/eclipse-score/docs-as-code/pull/481 first

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -157,22 +157,28 @@ def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good =
         deps = deps,
     )
 
+    # If the source directory is the root (".") we must omit it, otherwise:
+    # > invalid glob pattern './**/*.png': segment '.' not permitted
+    if source_dir == ".":
+        source_prefix = ""
+    else:
+        source_prefix = source_dir + "/"
+
     native.filegroup(
         name = "docs_sources",
         srcs = native.glob([
-            source_dir + "/**/*.png",
-            source_dir + "/**/*.svg",
-            source_dir + "/**/*.md",
-            source_dir + "/**/*.rst",
-            source_dir + "/**/*.html",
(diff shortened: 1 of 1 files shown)


docs.bzl
```diff
@@ -157,22 +157,28 @@ def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good =
         deps = deps,
     )
 
+    # If the source directory is the root (".") we must omit it, otherwise:
+    # > invalid glob pattern './**/*.png': segment '.' not permitted
```
MaximilianSoerenPollak, line 160: Does this not go against process recommendation / suggestion to have docs in 'docs' ?

AlexanderLanin: No idea. IMHO the tool should support not recommended usage, as long as we can easily enable it.

a-zw: It is a good practice because Sphinx scans the whole directory (including bazel-bin, .git, and whatnot) if you don't specify exclude_pattern in conf.py to avoid that. For example, it leads to confusing error messages like "file not included in any toctree" if there is any Markdown file not intended for the documentation.

AlexanderLanin: we have those in exclude_pattern: https://github.com/eclipse-score/docs-as-code/blob/b6d8d393a70ef13b29a5277fa14f897ee854d5bf/src/extensions/score_sphinx_bundle/__init__.py#L51

a-zw (changes requested):
Adapt documentation accordingly. At least how-to/setup and reference/bazel_macros

AlexanderLanin:
Ok, I'll do that in follow-up PR later today. If this one is technically fine, we can merge it, as to not block early users.

MaximilianSoerenPollak (approved):
Seems an alright solution for right now. 

Documentation Update will follow in next PR.

PR #483, opened 2026-04-13
Update score_process to 1.5.2

a-zw:
Bump dep version to latest.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -69,14 +69,4 @@ http_file(
     urls = ["https://github.com/withered-magic/starpls/releases/download/v0.1.21/starpls-linux-amd64"],
 )
 
-# Checker rule for CopyRight checks/fixes
-
-# docs dependency
-# Note: requirements were last aligned with 1.2.0,
-# the switch to 1.3.1 is purely to drop the dependency on docs-as-code 1.x.
-bazel_dep(name = "score_process", version = "1.5.1")
-git_override(
-    module_name = "score_process",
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -46,7 +46,7 @@ pip.parse(
 use_repo(pip, "docs_as_code_hub_env")
 
 # Additional Python rules provided by aspect, e.g. an improved version of
-bazel_dep(name = "aspect_rules_py", version = "1.4.0")
+bazel_dep(name = "aspect_rules_py", version = "1.11.2")
```
AlexanderLanin: please dont upgrade non dev dependencies at the moment, as we dont have an agreement yet about how to deal with the impact

```suggestion
bazel_dep(name = "aspect_rules_py", version = "1.4.0")
```

MaximilianSoerenPollak: Seems that we would need to do a bisect anyway to figure out what version is doing something different for us. 
As 1.11.2 does not work and breaks currently.

a-zw: Ok, reverted.

a-zw:
Expecting just consumer test score to fail with this warning:

    modules/os/operating_systems/docs/community/ebclfsa.rst:15: WARNING: comp__os_ebclfsa: is missing required link: `belongs_to`.


This is ok because the fix needs to be done in score repo anyways.

MaximilianSoerenPollak (approved):
Works with expected errors.

PR #484, opened 2026-04-13
feat: traceability metrics

### LINKED ISSUE #2774: Improvement: Introduce traceability coverage checker and CI reporting mode
Background
We already have a local traceability coverage checker that evaluates requirement evidence metrics (code links, test links, fully linked requirements, testcase linkage).
To support project-wide adoption, this checker should be integrated into docs-as-code in a controlled, low-risk rollout.

Goal
Introduce the checker as an official docs-as-code capability and run it in CI in reporting mode first, before strict gating.

Scope

Add the traceability coverage checker script to docs-as-code as maintained project functionality.
Ensure it can read generated needs.json in CI environments.
Add a CI step that always publishes the metric summary (reporting mode, non-blocking).
Document exact metric definitions and calculation scope in docs-as-code documentation.
Keep thresholds configurable by parameters, defaulting to non-blocking behavior for rollout.
Acceptance Criteria

Checker is part of the repository and executable via documented command.
CI runs checker and surfaces summary output on every relevant pipeline run.
Pipeline does not fail solely because of threshold misses in initial rollout mode.
Documentation explains each metric and input assumptions clearly.
A follow-up path to strict threshold gating is documented.
Out of Scope

New metamodel semantics or new cross-need link types.
Mandatory hard fail policy for all repos immediately.
Full AoU and architecture traceability enforcement.
Dependencies

### LINKED ISSUE #2775: Improvement: Align implementation dashboard with traceability coverage metrics
**Background**
Current dashboard views can be interpreted differently from CI traceability checker output, especially around implemented status versus evidence-backed traceability.
This creates confusion for project steering and adoption tracking.

**Goal**
Align dashboard semantics with the same metric logic used by the traceability coverage checker, while preserving a smooth transition.

**Scope**

Update dashboard sections to distinguish clearly between implementation status and traceability evidence status.
Reflect checker-relevant categories in dashboard views (for example code link coverage, test link coverage, fully linked coverage).
Add explanatory notes describing how dashboard and CI metrics relate.
Validate that dashboard interpretation and checker output are consistent on the same needs dataset.
Acceptance Criteria

Dashboard contains explicit evidence-based categories aligned with checker metrics.
Terminology is unambiguous (implemented versus documented/traceable).
Documentation explains metric mapping between dashboard and checker.
Example run shows no semantic contradiction between dashboard percentages and checker percentages.

**Out of Scope**

New mandatory metamodel checks.
Full rework of all requirement pages.
Repository-wide policy enforcement changes.

Relates to https://github.com/eclipse-score/score/issues/1586

FScholPer:
-> see https://github.com/eclipse-score/score/issues/2774


scripts_bazel/traceability_coverage.py
```diff
@@ -0,0 +1,542 @@
...
+    parser.add_argument(
+        "--min-req-code",
+        type=float,
+        default=0.0,
+        help="Minimum required percentage for requirements with source code links",
+    )
+    parser.add_argument(
+        "--min-req-test",
+        type=float,
+        default=0.0,
```
a-zw, line 261: Why not default=100.0?

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
 - Partial properties will lead to no Testlink creation.
   If you want a test to be linked, please ensure all requirement properties are provided.
 - Tests must be executed by Bazel first so `test.xml` files exist.
+
+
+CI/CD Gate for Linkage Percentage
```
a-zw, line 62: This is not really a How-To - that would cover how to add this to a workflow? Or maybe cover how to make a custom Bazel target with the script's json output?

Maybe this belongs to internals? Mostly, nobody will run this manually as it is intended for CI.

scripts_bazel/traceability_coverage.py
```diff
@@ -0,0 +1,542 @@
...
+        action="store_true",
+        help="Shortcut that enforces 100%% for all three minimum percentages",
+    )
+    parser.add_argument(
+        "--fail-on-broken-test-refs",
+        action="store_true",
+        help="Fail if a testcase references an unknown requirement ID",
+    )
+    parser.add_argument(
+        "--json-output",
```
a-zw, line 290: What is the json output for?

scripts_bazel/traceability_coverage.py
```diff
@@ -0,0 +1,542 @@
...
+
+def main() -> int:
+    parser = argparse.ArgumentParser(
+        description=(
+            "Compute requirement/test traceability coverage from sphinx-needs output "
+            "and optionally fail on threshold violations."
+        )
+    )
+    parser.add_argument(
+        "--needs-json",
```
a-zw, line 224: I think this should simply be mandatory and there is no need to "search" for needs.json.

a-zw (changes requested):

AlexanderLanin (changes requested):
see discussion in #485

docs/how-to/test_to_doc_links.rst
FScholPer: yeah but then we have to add it into cicid-workflows repo

scripts_bazel/traceability_coverage.py
FScholPer: Thats just if the path differs to increase flexibility

scripts_bazel/traceability_coverage.py
FScholPer: It generates a summary of the result as json. The parameter is just to change the path

FScholPer:
closed https://github.com/eclipse-score/docs-as-code/pull/485 @AlexanderLanin  can you please check again?

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+
+def is_non_empty(value: Any) -> bool:
+    """Return True if value should be treated as present for traceability checks."""
+    if value is None:
+        return False
+    if isinstance(value, str):
+        return bool(value.strip())
+    if isinstance(value, (list, tuple, set, dict)):
+        return len(value) > 0
+    return True
```
MaximilianSoerenPollak: Can simplify this via: 

```suggestion
def is_non_empty(value: Any) -> bool:
    """Return True if value should be treated as present for traceability checks."""
    if isinstance(value, str):
        return bool(value.strip())
    return bool(value)
```

As python already treats all other cases that you specified here the way you expect. 
Only 'special' one is that you want to not count whitespace, so we should keep that logic.

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+        return []
+    if isinstance(value, str):
+        return [item.strip() for item in value.split(",") if item.strip()]
+    if isinstance(value, list):
+        out: list[str] = []
+        for item in value:
+            if isinstance(item, str) and item.strip():
+                out.append(item.strip())
+        return out
+    return []
```
MaximilianSoerenPollak, line 45: Since you are already inside sphinx (due to it being inside docs-as-code extensions) you could also just get the all sphinx-needs from sphinx itself. 

You can do so via: 

```python
 Needs_Data = SphinxNeedsData(app.env)
 needs_data = set(x["id"] in Needs_Data.get_needs_view().values())
```
If you only wanted to have the ID's in the end. 

Though you then need to have this as a full extension. If you want I can convert this in a future PR to a full extension to enable this & other things.

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+) -> list[Any]:
+    """Extract requirements by type and implementation state."""
+    requirements: list[dict[str, Any]] = []
+    for need in all_needs:
+        need_type = str(need.get("type", "")).strip()
+        if need_type not in requirement_types:
+            continue
+        if not include_not_implemented:
+            implemented = str(need.get("implemented", "")).upper().strip()
+            if implemented not in {"YES", "PARTIAL"}:
```
MaximilianSoerenPollak, line 68: Implemented is tool_req specific. Unsure if this should be hidden / hardcoded in the code?

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+                continue
+        requirements.append(need)
+    return requirements
```
MaximilianSoerenPollak, line 71: If this was a sphinx-extension could also here make use of sphinx-needs internal functions etc. to get the needs list / objects for filtering etc. 
As mentioned above I can take this PR (after merge) and convert it in a future one.

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+        str(need.get("id", ""))
+        for need in requirements
+        if (
+            (
+                not is_non_empty(need.get("source_code_link"))
+                or not is_non_empty(need.get("testlink"))
+            )
+            and need.get("id")
+        )
+    ]
```
MaximilianSoerenPollak, line 110: Can be done in future PR: 

I would like to have this at least a tiny bit more efficient. 
Loops are one of the least efficient things in python, even though this already greatly improves due to use of set comprehension I think it should be possible with just one itteration through all requirements instead of 6. 

But as said, future PR improvement, for me currently not a block as we don't have 10k+ reqs to look through.

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+        "not_fully_linked_ids": sorted(not_fully_linked_ids),
+    }
+
+
+def calculate_test_metrics(
+    all_needs: Sequence[Any],
+    requirement_ids: set[str],
+    filtered_test_types: set[str],
+) -> dict[str, Any]:
+    """Calculate testcase linkage and broken testcase-reference statistics."""
```
MaximilianSoerenPollak, line 131: What exactly is meant by 'broken testcase-reference stats' ?

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+    ]
+    if filtered_test_types:
+        testcases = [
+            need
+            for need in testcases
+            if str(need.get("test_type", need.get("TestType", ""))).strip()
+            in filtered_test_types
+        ]
+
+    tests_total = len(testcases)
```
MaximilianSoerenPollak, line 143: This is not the total tests. This is the total tests that have had enough 'properties' and things to create a testcase need. 
This doesn't reflect the broken ones that didn't even come this far. These are all logged out to the console in the `source_code_linker` extension in the `xml_parser` file. 
There we already do this statistic.
Unsure if we should do it again, though also unsure if it's easy to get that out of that place.

Just wanted to make sure you know what exactly you will see here.

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+            if ref not in requirement_ids:
+                broken_references.append({"testcase": test_id, "missing_need": ref})
+
+    return {
+        "total": tests_total,
+        "filtered_test_types": sorted(filtered_test_types),
+        "linked_to_requirements": tests_linked,
+        "linked_to_requirements_pct": safe_percent(tests_linked, tests_total),
+        "broken_references": broken_references,
+    }
```
MaximilianSoerenPollak, line 168: Think this logic could be simpler via sphinx-needs internal extension things. 
Though that would be in a future Pr where this gets changed to a sphinx-extension

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def compute_traceability_summary(
+    all_needs: Sequence[Any],
+    requirement_types: set[str],
+    include_not_implemented: bool,
+    filtered_test_types: set[str],
+) -> dict[str, Any]:
+    """Return full CI/dashboard summary using one shared metric implementation."""
```
MaximilianSoerenPollak, line 177: Nit(not blocker for this PR): 

It's not really a full 'ci dashboard'. It's a dictionary with the statistics.

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
...
+
+    bazel test //...
+    bazel build //:needs_json
+    bazel run //scripts_bazel:traceability_coverage -- \
+       --needs-json bazel-bin/needs_json/_build/needs/needs.json \
+       --min-req-code 100 \
+       --min-req-test 100 \
+       --min-req-fully-linked 100 \
+       --min-tests-linked 100 \
+       --fail-on-broken-test-refs
```
MaximilianSoerenPollak, line 90: Does this work? 
I'm unsure if all our extensions run during `bazel build //:needs_json` if that is the case then ye this could work. 

Though question. Is there a specific `want` to be able to execute this as a *standalone* so to speak? 
Or would it be fine just running along side doc creation?

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
...
+
+To check only unit tests, filter testcase types:
+
+.. code-block:: bash
+
+    bazel run //scripts_bazel:traceability_coverage -- \
+  --needs-json bazel-bin/needs_json/_build/needs/needs.json \
+       --test-types unit-test
+
+Use lower thresholds during rollout and tighten towards 100% over time.
```
MaximilianSoerenPollak, line 117: We can also achieve this '100%' thing with two further changes, at least in some repos.

Make both source_code_link & testlink mandatory option for each requirement (par some exceptions). 

Make the pytest plugin  or sphinx extension throw an error if a test does not have properties => therefore isn''t linked

docs/internals/requirements/implementation_state.rst
```diff
@@ -20,12 +20,9 @@ Overview
 --------
 
 .. needpie:: Requirements Status
-   :labels: not implemented, implemented but not tested, implemented and tested
+   :labels: not implemented, implemented but incomplete docs, fully documented
```
MaximilianSoerenPollak, line 27: If I read this correctly, this seams like it does something different than before? 
statistics wise

scripts_bazel/tests/traceability_coverage_test.py
```diff
@@ -0,0 +1,233 @@
...
+import sys
+from pathlib import Path
+
+_MY_PATH = Path(__file__).parent
+
+
+def _write_needs_json(tmp_path: Path) -> Path:
+    needs_json = tmp_path / "needs.json"
+    payload = {
+        "current_version": "main",
```
MaximilianSoerenPollak, line 32: I don't think `current_version` can be a non ID (in a realistic scenario). 
Though don't think it matters.

scripts_bazel/tests/traceability_coverage_test.py
```diff
@@ -0,0 +1,233 @@
...
+                        "id": "TC_1",
+                        "type": "testcase",
+                        "partially_verifies": "REQ_1, REQ_2",
+                        "fully_verifies": "",
+                    },
+                    "TC_2": {
+                        "id": "TC_2",
+                        "type": "testcase",
+                        "partially_verifies": "",
+                        "fully_verifies": "",
```
MaximilianSoerenPollak, line 67: This would not be allowed.

If a test does not have the right prperties (either fully or partially verifies & other things) the testcase need will not be generated.

docs/reference/commands.md
```diff
@@ -5,6 +5,7 @@
 | `bazel run //:docs`                            | Builds documentation                                                                              |
 | `bazel run //:docs_check`                      | Verifies documentation correctness                                                                |
 | `bazel run //:docs_combo`                      | Builds combined documentation with all external dependencies included                             |
+| `bazel run //scripts_bazel:traceability_coverage -- --needs-json bazel-bin/needs_json/needs.json --min-req-code 100 --min-req-test 100 --min-req-fully-linked 100 --min-tests-linked 100 --fail-on-broken-test-refs` | Calculates requirement/test traceability percentages and fails if thresholds are not met |
```
a-zw: These commands are intended to work in any score repo. So it should be:


```suggestion
| `bazel run @score_docs_as_code//scripts_bazel:traceability_coverage -- --needs-json bazel-bin/needs_json/needs.json --min-req-code 100 --min-req-test 100 --min-req-fully-linked 100 --min-tests-linked 100 --fail-on-broken-test-refs` | Calculates requirement/test traceability percentages and fails if thresholds are not met |
```

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
...
+To enforce traceability in CI:
+
+1. Run tests.
+2. Generate ``needs.json``.
+3. Execute the traceability checker.
+
+.. code-block:: bash
+
+    bazel test //...
+    bazel build //:needs_json
```
a-zw, line 84: Makes no sense to me to run this build manually. The coverage check can simply depend on it and Bazel takes care of the execution.

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
...
+
+To enforce traceability in CI:
+
+1. Run tests.
+2. Generate ``needs.json``.
+3. Execute the traceability checker.
+
+.. code-block:: bash
+
+    bazel test //...
```
a-zw, line 83: We need a way to "build" a test report and then a coverage check can depend on that.

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
...
+    bazel build //:needs_json
+    bazel run //scripts_bazel:traceability_coverage -- \
+       --needs-json bazel-bin/needs_json/_build/needs/needs.json \
```
a-zw: Manually using files in bazel-bin is a red flag. Use Bazel dependencies instead.

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +53,43 @@ Limitations
...
 - Tests must be executed by Bazel first so `test.xml` files exist.
+
+
+CI/CD Gate for Linkage Percentage
+---------------------------------
+
+To enforce traceability in CI:
+
+1. Run tests.
+2. Generate ``needs.json``.
```
MaximilianSoerenPollak, line 71: This is a bit flawed, as currently testcase needs are generated as `external needs` and those are *excluded* from the needs.json. 

They can be included if we change the needs.json build filter which is possible so also external needs will be in there. 
It's like this in conf.py or probably in the score_metamodel extension: 

`needs_builder_filter = '(is_external==True and need_type==testcase)''`

The exact syntax we would have to figure out but something like this.

scripts_bazel/traceability_coverage.py
```diff
@@ -0,0 +1,401 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+"""Compute requirement and test traceability coverage from sphinx-needs output."""
```
a-zw, line 18: I believe this script, a CI quality gate, (and related files) should actually be in tooling and not in docs-as-code.

.gitignore
```diff
@@ -26,3 +26,4 @@ __pycache__/
 
 # bug: This file is created in repo root on test discovery.
 /consumer_test.log
+.clwb
```
a-zw, line 32: What are clwb files?

MaximilianSoerenPollak (commented):
Some comments for things I noticed. 

Will take another look after lunch.


Overall looks like genuine good work that we can build upon.

a-zw (changes requested):

MaximilianSoerenPollak:
Also a general thing. 

If there is any 'substantial' or partial generation from AI please add the disclaimer in the header. 
According to 'eclipse' something to the lines of this should be fine: 

```
# ╓                                                          ╖
# ║ Some portions generated by <AI Model Name>               ║
# ╙                                                          ╜

```

Model name meaning like 'Gpt' or 'Copilot' or 'Claude' etc.

scripts_bazel/traceability_coverage.py
FScholPer: would be to strict

FScholPer:
> Also a general thing.
> 
> If there is any 'substantial' or partial generation from AI please add the disclaimer in the header. According to 'eclipse' something to the lines of this should be fine:
> 
> ```
> # ╓                                                          ╖
> # ║ Some portions generated by <AI Model Name>               ║
> # ╙                                                          ╜
> ```
> 
> Model name meaning like 'Gpt' or 'Copilot' or 'Claude' etc.

I will add. But we should think about using tools like https://github.com/git-ai-project/git-ai in general

MaximilianSoerenPollak:
> I will add. But we should think about using tools like https://github.com/git-ai-project/git-ai in general

Probably might be worth a look to see if we could use that. I can bring it up in infra.

MaximilianSoerenPollak:
@FScholPer  I would propose that I can take these Ideas & Proposals that you have made in this PR and re-implement them into Docs-As-Code so they fit a better architecturally and makes sense with how testcases etc. actually behave & are created inside the DaC frameworks that deal with it. 

I fear it would be a lot more work to merge this PR and then rebuild it / change it quiet heavily later when we want to integrate it.


Do you think this would make sense and is a good way to move forward with this issue? 


@AlexanderLanin  @a-zw  please also advice here in your opinion what would be a good approach.

.gitignore
FScholPer: CLion IDE specific

src/extensions/score_metamodel/traceability_metrics.py
FScholPer: yeah that could be a way forward. But not for now

src/extensions/score_metamodel/traceability_metrics.py
FScholPer: refactored

src/extensions/score_metamodel/traceability_metrics.py
FScholPer: next pr plz

src/extensions/score_metamodel/traceability_metrics.py
FScholPer: fixed

scripts_bazel/traceability_coverage.py
FScholPer: You are right but for the moment as this is tightly coupled with docs-as-code I want it to be here(also not everyone uses tooling)

docs/internals/requirements/implementation_state.rst
FScholPer: fixed the wording to be align with the calculation

docs/how-to/test_to_doc_links.rst
FScholPer: Can you please create a separate issue? I don't want to mix it up now

scripts_bazel/tests/traceability_coverage_test.py
FScholPer: its a test

src/extensions/score_metamodel/traceability_metrics.py
FScholPer: future pr related

AlexanderLanin:
to me it sounds like we need to discuss the goals first, then how to achieve it. Thats the reason why this mixes extensions with custom executables and no-one really understands why and how it should look like.

docs/how-to/test_to_doc_links.rst
FScholPer: Something for future PR

docs/how-to/test_to_doc_links.rst
FScholPer: also future pr

FScholPer:
> to me it sounds like we need to discuss the goals first, then how to achieve it. Thats the reason why this mixes extensions with custom executables and no-one really understands why and how it should look like.

Goals for where 1. check linking of requirements to code/tests in ci 2. add dashboards in all modules. Or what goals do you have in mind?

scripts_bazel/traceability_coverage.py
a-zw: Sure, but what for? I see two use cases:

1. Generate an overview in the documentation.
2. CI quality gate.

Neither of them uses this json file.

If there is some use for it, then there should be a Bazel rule/target to consume it. However, it looks like this code should rather be removed because nobody needs it.

scripts_bazel/traceability_coverage.py
a-zw: I don't think it is a good idea to guess Bazel paths. Make it mandatory and explicit to loosen coupling between Script and Bazel.

AlexanderLanin:
@FScholPer we've discussed this PR, unfortunately we missed you. So here is our interpretation. Let's have a call next week.

The goals here are:
* unify statistics/metrics across modules
* provide thresholds for acceptable values, especially for CI
* provide a dedicated executable / bazel target??

Here is an idea, which simplifies the design:
image

What do you think? (May be not self explanatory)
Do you want to give that a try?

FScholPer:
> @FScholPer we've discussed this PR, unfortunately we missed you. So here is our interpretation. Let's have a call next week.
> 
> The goals here are:
> 
> * unify statistics/metrics across modules
> * provide thresholds for acceptable values, especially for CI
> * provide a dedicated executable / bazel target??
> 
> Here is an idea, which simplifies the design: image
> 
> What do you think? (May be not self explanatory) Do you want to give that a try?

I like the central json file to collect the results. I will check if this can be a way forward

MaximilianSoerenPollak:
> I like the central json file to collect the results. I will check if this can be a way forward


Let us know if there is any questions that arise or something that might be unclear, we gladly help answer them or clear up the confusion. 
Looking forward to seeing your work.

scripts_bazel/traceability_coverage.py
FScholPer: changed to the json approach from https://github.com/eclipse-score/docs-as-code/pull/484#issuecomment-4279365551

scripts_bazel/traceability_coverage.py
```diff
@@ -0,0 +1,414 @@
...
+import json
+import os
+import sys
+from pathlib import Path
+from typing import Any
+
+# Ensure shared metric code under src/ is importable when executed directly.
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
```
MaximilianSoerenPollak: This is not a good way to do this. 
We should not fiddle with the python path beyond bazel & python itself. 
It makes things very tricky to debug and even worse to understand.

scripts_bazel/traceability_coverage.py
```diff
@@ -0,0 +1,414 @@
+# *******************************************************************************
```
MaximilianSoerenPollak: I still do not see the appeal / need to execute this from OUTSIDE the documentation build directly. 
It makes the codebase a lot harder to not have access to internal sphinx things etc. 

I would strongly suggest to have this logic as a sphinx-needs extension instead of a stand alone script. 
If you need help or pointers with that conversion do not hesitate to reach out.

src/extensions/score_metamodel/sphinx_filters.py
```diff
@@ -0,0 +1,192 @@
...
+       :filter-func: score_metamodel.sphinx_filters.generic_pie_linked_items(std_req__mystandard__, gd_)
+
+"""
+
+from __future__ import annotations
+
+from sphinx_needs.need_item import NeedItem
+
+
+def generic_pie_linked_items(
```
MaximilianSoerenPollak: If it only checks the ''complies" link, then this is not generic. 

Could add a 3rd argument here to be the link that is suppose to be checked to make it truly generic.

src/extensions/score_metamodel/sphinx_filters.py
```diff
@@ -0,0 +1,192 @@
...
+    target_ids = [
+        str(n.get("id", ""))
+        for n in needs
+        if str(n.get("id", "")).startswith(id_prefix)
+    ]
+
+    linked_ids: set[str] = {
+        ref
+        for n in needs
+        if str(n.get("type", "")).startswith(compliance_prefix)
```
MaximilianSoerenPollak: Have to review the metamodel here to see if this is true in all cases. 
Theoretically we allow that the type is different from the mandated prefix.
So this might not work as you expect in those cases.

src/extensions/score_metamodel/sphinx_filters.py
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    connected = sum(1 for item_id in target_ids if item_id in linked_ids)
+    not_connected = len(target_ids) - connected
+
+    results.append(connected)
+    results.append(not_connected)
+
+
+def generic_pie_items_by_tag(
```
MaximilianSoerenPollak, line 91: Same as above, if you only check for the compliance link it's not generic.

src/extensions/score_metamodel/sphinx_filters.py
```diff
@@ -0,0 +1,192 @@
...
+    :filter-func: arguments:
+
+        - ``arg1`` – ID prefix of the work-product items to count
+          (e.g. ``std_wp__iso26262__``)
+        - ``arg2`` – type (or type prefix) of source needs whose ``complies``
+          lists are scanned (e.g. ``workproduct``)
+
+    Appends to *results*: ``[linked_count, not_linked_count]``
+    """
+    generic_pie_linked_items(needs, results, **kwargs)
```
MaximilianSoerenPollak: Not sure why this exists, seems like a useless abstraction? One could just call the `generic_pie_linked_items` directly.

src/extensions/score_metamodel/sphinx_filters.py
```diff
@@ -0,0 +1,192 @@
...
+    item_counts: dict[str, int] = {str(n.get("id", "")): 0 for n in items}
+
+    for container in containers:
+        for ref in container.get(field, []):
+            if ref in item_counts:
+                item_counts[ref] += 1
+
+    not_referenced = sum(1 for c in item_counts.values() if c == 0)
+    referenced_once = sum(1 for c in item_counts.values() if c == 1)
+    referenced_multiple = sum(1 for c in item_counts.values() if c > 1)
```
MaximilianSoerenPollak: Not important for this PR, but for future references, so only read if you are interested otherwise feel free to ignore. 

----


Python is rather bad at doing Arythmatic (simple additions etc.) can create over 100+ instruction cycles. 
I do not know why, it's stupid I agree. 

We are here only dealing with 2000-8000 items roughly. 
I have tested this (very briefly) with 25000 demo items `({"idx": 0})` and there the diference between the sum implementation here and something more optimized like this: 

`list(item_counts.values()).count()` is roughly 1,5seconds


Though as said this is just for information and optimizing for performance goes beyond the scope of this PR

src/extensions/score_metamodel/traceability_metrics.py
```diff
@@ -0,0 +1,258 @@
+# *******************************************************************************
```
MaximilianSoerenPollak, line 1: Same comment as earlier. 

This whole script can be made simpler / easier if it would not be standalone but a sphinx extension.

src/extensions/score_source_code_linker/tests/test_xml_parser.py
```diff
@@ -26,13 +26,14 @@
 import xml.etree.ElementTree as ET
 from collections.abc import Callable
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
```
MaximilianSoerenPollak: Here I'm unsure why the source code linker was touched at all. 

Why is it refactored & changed in this pr? 

In my opinion the refactor is not bad and seems to make sense in some way but this does not belong in this PR at all.

MaximilianSoerenPollak (dismissed):
Overall this is not bad work, thank you for this. 

Though there is some fundamental questions that I think need to be addressed. 

This goes a bit beyond / adrift from what we imagined this should be. 
I will try to make a bit clearer picture & describe it better so you can get a better idea of what exactly I mean. 
But before I go into those points, let me restate that this overall shows that you thought about the issue and put in work to provide something that fulfils a need an a issue we have been trying to solve. 
So do not feel discouraged by the comments and let me know if anything doesn't sound like it makes sense to you.

---


One point is that I do not see any reason that this should be standalone script able to execute standalone. 
The reason we suggested the JSON was so that the sphinx extension during docs build can generate the metrics json and save it. 
And then we can have a CLI script (in ci/cd workflows or wherever) that just reads this saved json and prints out the values & says yes, passed or no, not passed.

The reason to make this a Sphinx Extension is mainly due to simplicity of the code and speed, as well as information ability. 
When it runs inside sphinx during the docs build you have access (depending on when you connect) to all information you potentially need to make your calculations. 
You also have access to performance optimized functions to filter needs and other things that you then don't need to reimplement outside. 
Converting it to a Sphinx-Extension can be a bit daunting, so let us know if we can help with that task at all or if any questions arise. I gladly help any way I can.

The last point that I'm unsure about is the touching / changing of the Source Code Linker. 
The refactoring you did there seems to make sense to me in some ways, but I do not see any reason this should be in this PR at all, maybe you can enlighten me for the reason behind this.


--- 

A Picture, I hope this makes it a tiny bit clearer, please let me know if there are any questions.
image

scripts_bazel/traceability_coverage.py
FScholPer: removed

scripts_bazel/traceability_coverage.py
FScholPer: removed

src/extensions/score_metamodel/sphinx_filters.py
FScholPer: added 3rd argument and changed the complies. Please check again if it fits now

src/extensions/score_metamodel/sphinx_filters.py
FScholPer: fixed

src/extensions/score_metamodel/traceability_metrics.py
FScholPer: now the case

src/extensions/score_source_code_linker/tests/test_xml_parser.py
FScholPer: reverted

src/extensions/score_metamodel/sphinx_filters.py
FScholPer: replaced by list

scripts_bazel/traceability_coverage.py
FScholPer: removed entirely

FScholPer:
@AlexanderLanin can you please check again @MaximilianSoerenPollak can you please remove the blocking so that we can continue?

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,183 @@
...
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+Build Dashboards and Quality Gates
+==================================
+
+This guide is for repositories that *consume* docs-as-code as a Bazel
```
a-zw: No shit, Sherlock. This is in the how-to folder. No need to repeat obvious stuff.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,183 @@
...
+Run:
+
+.. code-block:: bash
+
+   bazel build //:needs_json
+
+This generates machine-readable output under:
+
+- ``bazel-bin/needs_json/_build/needs/needs.json``
+- ``bazel-bin/needs_json/_build/needs/metrics.json``
```
a-zw, line 66: The `:needs_json` target also provides metrics.json?? I don't see any corresponding changes in docs.bzl for that.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,183 @@
...
+   )
+
+In ``docs/conf.py``:
+
+.. code-block:: python
+
+   score_metamodel_requirement_types = "feat_req,comp_req,aou_req"
+   score_metamodel_include_external_needs = False
+
+Use ``score_metamodel_include_external_needs = True`` only in repositories that
```
a-zw, line 53: I don't like that this is configured via `conf.py`. Shouldn't we do that in Bazel somehow?

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,183 @@
...
+
+CI Quality Gate
+---------------
+
+After building ``//:needs_json``, run the gate on the exported metrics:
+
+.. code-block:: bash
+
+   bazel run @score_docs_as_code//scripts_bazel:traceability_gate -- \
+      --metrics-json bazel-bin/needs_json/_build/needs/metrics.json \
```
a-zw, line 112: Do we really need to specify metrics.json explicitly? Shouldn't there be a target which easily does the right thing?

docs/how-to/test_to_doc_links.rst
```diff
@@ -12,6 +12,10 @@
    # SPDX-License-Identifier: Apache-2.0
    # *******************************************************************************
 
+   # ╓                                                          ╖
+   # ║ Some portions generated by Github Copilot                ║
+   # ╙                                                          ╜
+
```
a-zw: Why add this here?

docs/how-to/test_to_doc_links.rst
```diff
@@ -53,3 +57,117 @@ Limitations
...
+2. ``traceability_gate`` reads that ``metrics.json`` and enforces configurable
+   coverage thresholds.
+
+Separating computation (Sphinx extension, during docs build) from gating (thin
+CLI, in CI) keeps the gate decoupled from the Sphinx/Bazel build: it never
+parses ``needs.json`` itself and has direct access to all sphinx-needs data.
+
+.. note::
+
+   ``metrics.json`` is the **single source of truth** for traceability data.
```
a-zw: Since it is so essential, it should have its own target maybe?

docs/internals/requirements/implementation_state.rst
```diff
@@ -11,21 +11,60 @@
    #
    # SPDX-License-Identifier: Apache-2.0
    # *******************************************************************************
+
+   # ╓                                                          ╖
+   # ║ Some portions generated by Github Copilot                ║
+   # ╙                                                          ╜
```
a-zw:
```suggestion
```

Did you explicitly ask for this or does Copilot do this on its own?

a-zw (dismissed):

docs/internals/requirements/implementation_state.rst
MaximilianSoerenPollak: This disclaimer is one of the ways that we convey that AI was used in some portion. 
The other 'acceptable' way was to put a disclaimer around exactly what was generated / changed.

FScholPer: https://www.eclipse.org/projects/handbook/#genai

a-zw: Then we should really go by their recommendation:

> Using Assisted-by tags is emerging as a de facto standard for capturing the identity of an AI assistant in a machine-readable format.  

```suggestion

   #  Assisted-by: GitHub Copilot
```

Though it seems like the actual idea is to put it at the top of the file like SPDX-License-Identifier.

docs/how-to/dashboards_and_quality_gates.rst
FScholPer: I think a small section about what target group this dashboard taggle is not bad

FScholPer:
Fixed the new comments. Mostly documentation issues

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,211 @@
...
+           "@score_process//:needs_json",
+       ],
+       scan_code = [":module_sources"],
+   )
+
+In ``docs/conf.py``:
+
+.. code-block:: python
+
+   score_metamodel_requirement_types = "feat_req,comp_req,aou_req"
```
AlexanderLanin, line 50: requirements types are already flagged in metamodel as such

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,211 @@
...
+   score_metamodel_include_external_needs = False
```
AlexanderLanin, line 51: naming it `x` and explaining that it does `y` seems complicated ;-)
Let's name it `aggregate_traceability_across_dependencies`

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,211 @@
...
+
+With the ``docs(...)`` macro and ``score_metamodel`` extension enabled, your
+repository can:
+
+- build an HTML dashboard from its own Sphinx needs,
+- include external needs from other repositories when desired,
+- export ``needs.json`` and ``metrics.json`` for machine-readable reporting,
+- gate CI on traceability thresholds via ``traceability_gate``.
+
+Typical Setup
```
AlexanderLanin, line 38: this repeats quite a lot of the setup documentation. Let's refer to :ref:`setup`. instead of repeating the entire setup.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,211 @@
...
+
+- Set ``score_metamodel_include_external_needs = True``.
+- Aggregate requirements across module dependencies when that is the intended
+  repository purpose.
+- Use this for system or integration-level dashboards.
+
+CI Quality Gate
+---------------
+
+After building ``//:needs_json``, run the gate on the exported metrics:
```
AlexanderLanin: After building/running any docs command (i.e. `bazel build //:needs_json` or `bazel run //:docs_verify` are the fastest ones).

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,211 @@
...
+
+.. code-block:: bash
+
+   bazel run @score_docs_as_code//scripts_bazel:traceability_gate -- \
```
AlexanderLanin: let's hide `scripts_bazel` from this command

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,211 @@
...
+In CI, wire targets through Bazel dependencies so test execution and
+``needs_json`` generation happen before the gate target.
+
+In larger repositories, define a dedicated wrapper target for your standard
+gate thresholds so CI calls a single Bazel target.
+
+Useful flags:
+
+- ``--require-all-links`` for strict 100 percent gating
+- ``--fail-on-broken-test-refs`` to fail when testcase references point to
```
AlexanderLanin: isnt that always the case, and not "disableable"?

docs/internals/requirements/implementation_state.rst
```diff
@@ -11,21 +11,58 @@
...
+
+This page shows how the docs-as-code tooling covers process and tool
+requirements. It focuses on tooling capabilities offered to downstream
+repositories rather than on product-specific traceability inside those
+repositories.
 
 Overview
 --------
 
 .. needpie:: Requirements Status
```
AlexanderLanin:
```suggestion
.. needpie:: Tool Requirements Status
```

docs/internals/requirements/implementation_state.rst
```diff
@@ -11,21 +11,58 @@
...
+
+The overview pie combines implementation state and traceability evidence:
+
+- ``not implemented``:
+   requirement has ``implemented == NO``.
+- ``implemented but incomplete traceability``:
+   requirement has ``implemented == YES`` or ``implemented == PARTIAL``,
+   but is missing at least one traceability link (code link and/or test link).
+- ``fully linked``:
+   requirement has both ``source_code_link`` and ``testlink``.
```
AlexanderLanin: and implemented?!

docs/internals/requirements/implementation_state.rst
```diff
@@ -11,21 +11,58 @@
...
-   type == 'tool_req' and testlink != '' and (implemented == 'YES' or implemented == 'PARTIAL')
+Implementation labels used on this page:
+
+- ``NO``: requirement is not implemented.
+- ``PARTIAL``: requirement is partly implemented.
+- ``YES``: requirement is implemented.
+
+Why multiple pies are shown:
+
+- ``Requirements with Codelinks`` shows implementation-to-source traceability.
```
AlexanderLanin:
```suggestion
- ``Requirements with Codelinks`` shows requirement-to-implementation traceability.
```

docs/internals/requirements/implementation_state.rst
```diff
@@ -11,21 +11,58 @@
...
+- ``Requirements with linked tests`` shows implementation-to-verification traceability.
```
AlexanderLanin:
```suggestion
- ``Requirements with linked tests`` shows requirement-to-verification traceability.
```

docs/internals/requirements/index.rst
```diff
@@ -15,11 +15,29 @@
 Requirements
 ============
 
+This repository provides the docs-as-code tooling used by other SCORE
+repositories. The pages in this section therefore focus on two questions:
+
+1. Which process and tool requirements are covered by the docs-as-code tooling?
```
AlexanderLanin:
```suggestion
1. Which process requirements are covered by the docs-as-code tooling?
```

docs/reference/commands.md
```diff
@@ -1,10 +1,17 @@
...
+# ║ Some portions generated by Github Copilot                ║
+# ╙                                                          ╜
+-->
+
 # Commands
 
 | Target                                         | What it does                                                                                      |
 | ---------------------------------------------- | ------------------------------------------------------------------------------------------------- |
-| `bazel run //:docs`                            | Builds documentation                                                                              |
+| `bazel run //:docs`                            | Builds documentation (also writes `metrics.json` via the score_metamodel extension)               |
```
AlexanderLanin: Users dont need to know about our internal components

```suggestion
| `bazel run //:docs`                            | Builds documentation (also writes `metrics.json`)               |
```

src/extensions/score_source_code_linker/__init__.py
```diff
@@ -104,9 +107,15 @@ def build_and_save_combined_file(outdir: Path):
...
             source_code_links_json
         )
-    test_code_links = load_test_xml_parsed_json(
-        get_cache_filename(outdir, "score_xml_parser_cache.json")
-    )
+    test_cache = get_cache_filename(outdir, "score_xml_parser_cache.json")
+    if test_cache.exists():
+        test_code_links = load_test_xml_parsed_json(test_cache)
+    else:
+        LOGGER.debug(
```
AlexanderLanin, line 114: how can this happen?

src/extensions/score_source_code_linker/__init__.py
```diff
@@ -277,14 +301,13 @@ def setup_once(app: Sphinx):
     )
     LOGGER.debug(f"DEBUG: Git root is {find_git_root()}")
 
-    # Run only for local files!
-    # ws_root is not set when running on external repositories (dependencies).
+    # Run for local files if possible. In Bazel sandbox builds, ws_root may be
```
AlexanderLanin, line 304: ⚠️ 
this is quite a logic change, as it will now run stuff on internal and external needs_json invocations. Probably biggest impact in this small looking if statement! Any impact? I need to consider consequences.

src/extensions/score_source_code_linker/__init__.py
```diff
@@ -300,6 +323,13 @@ def setup(app: Sphinx) -> dict[str, str | bool]:
     # Esbonio will execute setup() on every iteration.
     # setup_once will only be called once.
     app.add_config_value("KNOWN_GOOD_JSON", default="", rebuild="env", types=str)
+    app.add_config_value("score_sourcelinks_json", default="", rebuild="env", types=str)
+    app.add_config_value(
+        "score_source_code_linker_plain_links",
```
AlexanderLanin: why two new config values?

src/extensions/score_source_code_linker/__init__.py
```diff
@@ -316,6 +346,102 @@ def find_need(all_needs: NeedsMutable, id: str) -> NeedItem | None:
...
+        if need.get("testlink"):
+            LOGGER.debug(
+                f"?? Need {need_id} already has testlink: {need.get('testlink')}"
+            )
+
+
+def _render_code_link(plain_links: bool, metadata: RepoInfo, link: NeedLink) -> str:
+    if plain_links:
+        return (
+            "https://github.com/placeholder/placeholder/blob/unknown/"
```
AlexanderLanin: why? 😆

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: found the code change. I would not add this?! We also don't have a flag to ignore broken requriements -> requirements links, so why should we ignore broken test -> requirements links?

docs/how-to/dashboards_and_quality_gates.rst
FScholPer: sure but this is an optional customization for the individual repos which want to build their own config. So remove or not?

docs/reference/commands.md
```diff
@@ -1,10 +1,17 @@
...
 | `bazel run //:docs_check`                      | Verifies documentation correctness                                                                |
 | `bazel run //:docs_combo`                      | Builds combined documentation with all external dependencies included                             |
+| `bazel run //:traceability_gate -- --metrics-json bazel-bin/needs_json/_build/needs/metrics.json --min-req-code 70 --min-req-test 70 --min-req-fully-linked 60 --min-tests-linked 70` | Reads the pre-computed metrics.json and fails if coverage thresholds are not met                   |
```
AlexanderLanin, line 14: _AI-assisted review (Claude):_

`//:traceability_gate` is documented here but no such target is defined in `BUILD` or created by the `docs()` macro in `docs.bzl`. Only `//scripts_bazel:traceability_gate` exists.

For the docs-as-code repo itself, add an alias in `BUILD`. For consumer repos, the `docs()` macro should create this target automatically — similar to how it creates `//:docs_check` and `//:needs_json`.

**Must fix:** The documented command won't work for anyone.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -0,0 +1,142 @@
...
+
+CI Quality Gate
+---------------
+
+After building ``//:needs_json``, run the gate on the exported metrics:
+
+.. code-block:: bash
+
+   bazel run //:docs && \
+   bazel run //:traceability_gate -- \
```
AlexanderLanin, line 111: _AI-assisted review (Claude):_

Same issue: `bazel run //:traceability_gate` won't work for consumer repos because the `docs()` macro doesn't create this target. The how-to should also clarify the dependency: the gate needs `metrics.json` which is a side-effect of the Sphinx build, but Bazel doesn't declare it as an output — so `bazel build //:needs_json && bazel run //:traceability_gate` may not find the file.

Consider wiring the gate as a Bazel test target that depends on `:needs_json`, so `bazel test //:traceability_gate` does the right thing automatically.

scripts_bazel/BUILD
```diff
@@ -37,3 +41,11 @@ py_binary(
...
     visibility = ["//visibility:public"],
 )
+
+py_binary(
+    name = "traceability_gate",
+    srcs = ["traceability_gate.py"],
+    main = "traceability_gate.py",
+    visibility = ["//visibility:public"],
+    deps = all_requirements,
+)
```
AlexanderLanin, line 47: _AI-assisted review (Claude):_

`traceability_gate.py` only uses stdlib modules (`argparse`, `json`, `os`, `sys`, `pathlib`). `deps = all_requirements` adds ~30 pip packages unnecessarily.

```suggestion
    deps = [],
```

src/extensions/score_source_code_linker/__init__.py
```diff
@@ -316,6 +365,102 @@ def find_need(all_needs: NeedsMutable, id: str) -> NeedItem | None:
...
```
AlexanderLanin: _AI-assisted review (Claude):_

This hardcoded placeholder URL will end up in `needs.json` and could propagate into HTML when needs are imported as external needs by consuming repos. Users clicking these links would land on a 404.

Since `plain_links` mode is for sandbox builds where git info is unavailable, consider dropping the URL entirely and using just the display name:
```python
return f"{link.file}:{link.line}"
```
The `source_code_linker` string-link regex already handles the name-only format (no `<>` separator).

AlexanderLanin (commented):
_Note: This review was AI-assisted (Claude)._

---

## Summary

The architectural direction of this PR is solid and the iteration since the early feedback shows real progress: metrics are now computed inside Sphinx (single source of truth), exported as a well-defined `metrics.json` (schema-v1), and consumed by a clean gate script. The shared `traceability_metrics.py` module avoids duplication between dashboards and CI. The refactoring of `standards.py` to use `generic_pie_*` filters is a nice generalization that will pay off as more standards are onboarded. Test coverage for the new modules is good.

I'm organizing findings into three tiers so we can focus the remaining effort.

### Must fix before merge

1. **`//:traceability_gate` target does not exist for consumer repos.** The `docs()` macro in `docs.bzl` does not create a `traceability_gate` target. The how-to guide (line 108) and `commands.md` both reference `bazel run //:traceability_gate`, but only `//scripts_bazel:traceability_gate` exists. Consumer repos that call `docs()` will get `docs`, `docs_check`, `needs_json`, etc. — but not `traceability_gate`. The macro should create an alias or wrapper so the documented command works out of the box. (See inline comments on `commands.md` and the how-to.)

2. **`metrics.json` is an undeclared Bazel output.** `_write_metrics_json` writes to `Path(app.outdir) / "metrics.json"` as a side effect of the Sphinx build, but `docs.bzl` doesn't declare it as an output of the `:needs_json` target. For `bazel run` this may work because the sandbox is relaxed, but for `bazel build` (which CI typically uses) the file may not be preserved or accessible at the documented path. This was flagged by @a-zw as well — needs explicit Bazel integration.

### Should fix

3. **Placeholder GitHub URL in plain_links mode.** `_render_code_link` generates `https://github.com/placeholder/placeholder/blob/unknown/…` for Bazel sandbox builds. These fake URLs could propagate into `needs.json` and eventually into external-needs HTML. Consider using just the local path without a URL (the `<>` separator format already supports name-only rendering), or at minimum a clearly-sentinel scheme like `local://`.

4. **Unrelated changes inflate the diff.** The `.pre-commit-config.yaml` switch to `tools/run_tool.sh` wrappers, the `tools/run_tool.sh` script itself, and the shellcheck/actionlint aliases in `BUILD` are orthogonal to traceability metrics. Splitting them into a separate PR would make both easier to review and revert independently.

5. **`traceability_gate` Bazel target pulls `all_requirements` but only uses stdlib.** The gate script imports only `argparse`, `json`, `os`, `sys`, `pathlib`. The `deps = all_requirements` in `scripts_bazel/BUILD` adds ~30 pip packages for no reason, slowing down resolution and increasing the dependency surface.

### Can be fixed later

6. **`process_requirements` computed but not exported to `metrics.json` / gate.** `compute_traceability_summary` returns process-requirement metrics and the dashboard shows them, but `_write_metrics_json` doesn't include them in the JSON output and the gate can't enforce thresholds on them. Fine for v1, but worth noting as a gap for future iterations.

7. **Global mutable `_DEFAULT_INCLUDE_EXTERNAL` in `traceability_dashboard.py`.** Using module-level state with `set_default_include_external()` works but is fragile for testing and any future multi-build scenario. A config-object approach would be cleaner.

8. **CSS `!important` in `score_design.css`.** `box-shadow: none !important` on `.score-grid .sd-card` overrides any consuming repo's card styling. This is fine for now but may need scoping to dashboard-specific containers later.

### What's good

- The metrics.json schema is well-designed: versioned, documented, and strict (`additionalProperties: false`). This will make evolution manageable.
- Reusing `compute_traceability_summary` across dashboard filters and JSON export guarantees consistency — the numbers you see in the pie charts are the same ones CI gates on.
- The gate script is clean and focused: it reads pre-computed data, prints a human-readable summary, and exits with distinct codes (0/1/2). Easy to integrate into any CI system.
- Test coverage for the new code is good, especially the `test_traceability_metrics.py` and `traceability_gate_test.py` suites which cover edge cases like external needs, multiple types, and broken references.
- The how-to guide (`dashboards_and_quality_gates.rst`) reads well and covers the right workflow: local → aggregated → CI rollout.

docs/reference/commands.md
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `0d9195cb`: `docs()` macro now creates a `traceability_gate` alias target for consumer repos.

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `0d9195cb`: `docs()` macro now creates a `traceability_gate` alias. The how-to has also been updated to clarify the dependency on `metrics.json` from the Sphinx build (`69254cfe`).

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `0d9195cb`: `docs()` macro now exposes `//:traceability_gate` for all consumer repos.

scripts_bazel/BUILD
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `44dcc749`: `deps` changed to `[]` — the gate script only uses stdlib modules.

src/extensions/score_source_code_linker/__init__.py
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `cba4fa8f`: placeholder URL removed. Plain-links mode now renders as `file:line` without a fake GitHub URL.

src/extensions/score_source_code_linker/__init__.py
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `cba4fa8f`: removed the hardcoded placeholder URL entirely. Plain-links mode now uses just the local path.

docs/internals/requirements/implementation_state.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

AI disclaimers removed from non-source files (`ec150c86`, `85a9d261`, `fa04eda0`). Source files that retain the tag now use `# Assisted-by: GitHub Copilot` format as suggested.

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `69254cfe`: how-to now states that any docs build writes `metrics.json` and shows the correct paths.

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `69254cfe`: documentation now explains the relationship — `metrics.json` is written by the Sphinx build (score_metamodel extension) and lands in the `needs_json` output directory.

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `89d2ff08`: intro rewritten to a concise purpose statement.

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `89d2ff08`: setup section now refers to `:ref:\`setup\`` instead of repeating the documentation.

docs/reference/commands.md
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `18d2c0d6`: simplified to `Builds documentation (also writes metrics.json)` without mentioning the internal extension name.

docs/internals/requirements/implementation_state.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `fb202b4a`: "fully linked" now requires implementation — wording clarifies that a requirement must be implemented and have both `source_code_link` and `testlink`.

docs/internals/requirements/implementation_state.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `fb202b4a`: wording updated to `Requirements with Codelinks`.

docs/internals/requirements/implementation_state.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `fb202b4a`: wording updated to `Requirements with linked tests`.

src/extensions/score_source_code_linker/__init__.py
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `be08bfce`: corrected the debug log guard (condition was inverted).

docs/internals/requirements/implementation_state.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `fb202b4a`: needpie title changed to `Tool Requirements Status`.

docs/how-to/dashboards_and_quality_gates.rst
AlexanderLanin: _Fix-commit created by AI (Claude)_

Fixed in `89d2ff08` and `0d9195cb`: all user-facing commands now use `//:` targets. `scripts_bazel` paths are only shown in the internal targets section.

AlexanderLanin:
## Review fixes — batch update (AI-assisted, Claude)

Pushed 14 commits addressing findings from the May 11–12 review round. Individual reply comments have been added to each resolved thread.

### Fixes

**Bazel integration**
- `docs()` macro now creates `//:traceability_gate` alias for consumer repos (`0d9195cb`)
- Removed unnecessary `all_requirements` pip deps from `traceability_gate` target — script uses only stdlib (`44dcc749`)

**Documentation cleanup**
- How-to intro rewritten to concise purpose statement; setup section refers to `:ref:`setup`` instead of repeating (`89d2ff08`)
- Corrected nonexistent `docs_verify` reference; cleaned up wording (`89d2ff08`)
- Clarified that any docs build writes `metrics.json`; documented correct artifact paths (`69254cfe`)
- `commands.md`: simplified docs description, hid internal extension name; user-facing targets use `//:` prefix (`18d2c0d6`)

**Implementation state page**
- "Fully linked" now requires implementation (not just links); wording updated per suggestions (`fb202b4a`)
- Needpie title → `Tool Requirements Status`; description labels updated (`fb202b4a`)

**Source code linker**
- Removed hardcoded `placeholder/placeholder` GitHub URL from plain-links mode — now renders `file:line` only (`cba4fa8f`)
- Fixed inverted debug log guard in `_log_existing_links` (`be08bfce`)

**Housekeeping**
- AI disclaimers removed from `.gitignore`, `scripts_bazel/BUILD`, `scripts_bazel/tests/BUILD` (`ec150c86`, `85a9d261`, `fa04eda0`)
- Added missing `include_external` field to metrics JSON schema + gate test fixtures (`2b4e5c4e`, `2f65d264`)
- Added missing `known_good` parameter to `docs()` macro table (`74ddf9b6`)

### Still open / not addressed here

- **`requirement_types` config** — still documented as user-facing `conf.py` setting (AlexanderLanin flagged that types are already in metamodel; FScholPer asked whether to remove). Needs decision.
- **Two config values** `KNOWN_GOOD_JSON` vs `score_sourcelinks_json` — purpose overlap unclear (AlexanderLanin's question unanswered).
- **`AssertionError` control flow** — acknowledged tech debt, not a blocker.
- **Source code linker logic change** (runs on both internal/external `needs_json`) — documented with rationale but impact assessment still pending.

AlexanderLanin:
Here are all my changes on one screen, please have a look: https://github.com/eclipse-score/docs-as-code/pull/484/changes/fcde17875d5ef3b72366b9ac61fa2a03ded36123..389675d6e6845eecf9593649b05dc4a303bc5f5b

FScholPer:
> Here are all my changes on one screen, please have a look: https://github.com/eclipse-score/docs-as-code/pull/484/changes/fcde17875d5ef3b72366b9ac61fa2a03ded36123..389675d6e6845eecf9593649b05dc4a303bc5f5b

Looks good for me

AlexanderLanin (approved):

PR #485, opened 2026-04-13
docs(traceability): align dashboards with coverage metrics

FScholPer:
--> see https://github.com/eclipse-score/score/issues/2775


docs/internals/requirements/requirements.rst
```diff
@@ -731,6 +731,22 @@ Architecture Attributes
...
+.. needpie:: Tool Requirements with Source Code Links
+  :labels: with codelink, without codelink
+  :colors: green, red
+
+  type == 'tool_req' and source_code_link != ''
+  type == 'tool_req' and source_code_link == ''
+
+.. note::
+  This dashboard is complemented by an automated CI check using
+  ``traceability_coverage``.
```
AlexanderLanin, line 745: I wonder whether its easier / more convenient to extend those diagrams in some clever way. More effort... but otherwise we will always have huge discrepancies between thresholds and dashboards. That sounds like a maintenance nightmare.

```
.. needpie:: Tool Requirements with Source Code Links
  :labels: with codelink, without codelink
  :colors: green, red
  :minimum: 80, 0

  type == 'tool_req' and source_code_link != ''
  type == 'tool_req' and source_code_link == ''
```

AlexanderLanin: Idea discussed with @MaximilianSoerenPollak. We can probably indeed solve it "like that". This avoids having thresholds and metrics showing different values.

AlexanderLanin (changes requested):
see comment above

docs/internals/requirements/requirements.rst
FScholPer: okay so how to go on then?

MaximilianSoerenPollak: We had the idea to solve it in ref-integration & the consumers.

The integration in re-fint would give you the dashboard for each repo that you wanted. 
And the implementation  in the consumers (like docs-as-code etc.) would give those repos a statistic as well in their own documentation. 

The integration itself would be some thing like a 'generic' needpie function that can be called form the needpie via 'filter-func' attribute. 
And then it calculates the things. 
This way we have a dashboard and the statistic itself inside the docs therefore do not have risk to show different values. 


If you want, we can do this and implement it, or you can give it a go as well. Just let us know.

FScholPer: yeah but there should be also a dashboard in each of the modules as a must have as they are SEooC. This goes in the same direction as the DR-008

MaximilianSoerenPollak: Yes that's why there is a duplication of this. 
The code will be in ref-integration and in each of the modules. So they both have these dashboards / statistics.

FScholPer: hm okay but then is there a way of just link it so that we have it in one central place ?

MaximilianSoerenPollak: Most likely not as the code that needpie needs to reference hast to be in the same repository. 

Though I will see if this is possible to maybe link somehow with a known path etc. If that is possible, then yes this would be the preferred way.

FScholPer: I can try to implement that generic filter-func and ping you until its available

FScholPer:
Closed as I removed the dashboard and add public filters to 2774

PR #501, opened 2026-04-20
chore: drop local score_tooling dependency

AlexanderLanin:
In order to try out renovate-bot, we need clean builds which are not possible as long as tooling is involved.


.pre-commit-config.yaml
```diff
@@ -27,6 +27,15 @@ repos:
...
         args: ["--maxkb=150"]
 
+  - repo: local
+    hooks:
+      - id: bazel-mod-tidy
+        name: format Bazel module files
+        entry: bazel mod tidy
+        language: system
+        pass_filenames: false
+        files: '(^|/)(MODULE\.bazel|.*\.bzl|BUILD(\.bazel)?)$'
```
MaximilianSoerenPollak, line 37: I already hate this regex. Is this the right way? 
But if it works I guess it makes sense.

MaximilianSoerenPollak (approved):
Seems fine to me though. 
With this we now should have no tooling dependency (direclty in DaC) 
Once Process releases again it should then be even in consumers gone.

PR #537, opened 2026-05-18
Fix: Fix rst based tests

MaximilianSoerenPollak:
Expect-Not tests were not executed correctly and often were a no-op. 
This change fixes that and makes them actually test things and relevant.

Changed behaviour of test checking too to enable more general usage.
Also changed all of the EXPECT-NOT to be generic instead of specific.

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/extensions/score_metamodel/tests/rst/graph/test_metamodel_graph.rst
+++ b/src/extensions/score_metamodel/tests/rst/graph/test_metamodel_graph.rst
@@ -31,7 +31,7 @@
 
 
 .. Positive Test: Child requirement QM. Parent requirement has the correct related safety level. Parent requirement is `QM`.
-#EXPECT-NOT: feat_req__child__1: Parent need `feat_req__parent__QM` does not fulfill condition `safety == QM`. Explanation: An ASIL requirement must link at least one parent/upstream ASIL requirement for correct decomposition. Please ensure the parent’s safety level is QM and its status is valid.
+#EXPECT-NOT: safety
 
 .. feat_req:: Child requirement 1
    :id: feat_req__child__1
@@ -41,7 +41,7 @@
 
 
 .. Positive Test: Child requirement ASIL B. Parent requirement has the correct related safety level. Parent requirement is `QM`.
-#EXPECT-NOT: feat_req__child__2: Parent need `feat_req__parent__ASIL_B` does not fulfill condition `safety == QM`. Explanation: An ASIL requirement must link at least one parent/upstream ASIL requirement for correct decomposition. Please ensure the parent’s safety level is QM and its status is valid.
+#EXPECT-NOT: safety
 
 .. feat_req:: Child requirement 2
    :id: feat_req__child__2
@@ -62,17 +62,6 @@
 
 
 
-.. Parent requirement does not exist
-#EXPECT: unknown outgoing link
-
-.. feat_req:: Child requirement 4
-   :id: feat_req__linking_to_unknown_parent
-   :safety: ASIL_B
-   :status: valid
-   :satisfies: feat_req__parent0__abcd
-
-
-
 .. Mitigation of Safety Analysis (FMEA and DFA) shall be checked. Mitigation shall have the same or higher safety level than the analysed item.
 .. Negative Test: Linked to a mitigation that is lower than the safety level of the analysed item.
 #EXPECT: feat_saf_dfa__child__5: Parent need `feat_req__parent__QM` does not fulfill condition `safety != QM`. Explanation: An ASIL_B safety requirement must link to a ASIL_B requirement. Please ensure that the linked requirements safety level is not QM and it's status is valid.
@@ -85,7 +74,7 @@
 
 
 .. Positive Test: Linked to a mitigation that is equal to the safety level of the analysed item.
-#EXPECT-NOT: feat_saf_dfa__child__6: Parent need `feat_req__parent__ASIL_B` does not fulfill condition `safety != QM`. Explanation: An ASIL_B safety requirement must link to a ASIL_B requirement. Please ensure that the linked requirements safety level is not QM and it's status is valid.
+#EXPECT-NOT: safety
 
 .. feat_saf_dfa:: Child requirement 6
    :id: feat_saf_dfa__child__6
@@ -106,7 +95,7 @@
 
 
 .. Positive Test: Linked to a mitigation that is equal to the safety level of the analysed item.
-#EXPECT-NOT: comp_saf_dfa__child__8: Parent need `feat_req__parent__ASIL_B` does not fulfill condition `safety != QM`. Explanation: An ASIL_B safety requirement must link to a ASIL_B requirement. Please ensure that the linked requirements safety level is not QM and it's status is valid.
+#EXPECT-NOT: safety
 
 .. comp_saf_dfa:: Child requirement 8
    :id: comp_saf_dfa__child__8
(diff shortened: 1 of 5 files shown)


AlexanderLanin (commented):

**Critical: 0 | Important: 2 | Suggestions: 2**

The core fix — switching `filter_warnings_by_position` from substring-matching to `startswith`-with-stripping — is correct and the ANSI-code scrubbing is a good catch. Two issues need attention before merging.

**Key concerns:**

1. Pre-filtering `warnings` to lines containing `"score_metamodel"` silently makes it impossible to assert on warnings from other Sphinx extensions (e.g., sphinx-needs itself). The one existing `#EXPECT: unknown outgoing link` test was a casualty of this — the warning comes from sphinx-needs, not score_metamodel, so it disappears before any matching occurs. The test was removed rather than the root cause being documented or addressed.

2. The `std_wp__test__abcd` EXPECT-NOT in `test_id_contains_feature.rst` was removed. If this EXPECT-NOT was previously a no-op (as the PR description says), and now fails after the fix, it means the "explicitly allowed" behavior is actually broken in production. Removing the test hides a real regression rather than fixing it.

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -192,8 +207,13 @@ def test_rst_files(
     app.build()
```
AlexanderLanin, line 208: _AI-assisted review (Claude):_

**Important:** Pre-filtering `warnings` to only `"score_metamodel"` lines is a semantic change that narrows the test scope silently. The removed `#EXPECT: unknown outgoing link` test (in `test_metamodel_graph.rst`) is a direct casualty: that warning originates from sphinx-needs, not score_metamodel, so it will never appear in this list. Any EXPECT annotation on a non-score_metamodel warning will now silently pass as a false positive.

If the intent is to restrict the test harness to score_metamodel warnings only, document that constraint clearly (e.g., a comment or docstring on the filter line), and explicitly state in the PR description that non-score_metamodel assertions are out of scope. The removed test should either be restored with a mechanism to include sphinx-needs warnings, or documented as intentionally dropped coverage.

src/extensions/score_metamodel/tests/rst/graph/test_metamodel_graph.rst
```diff
@@ -62,17 +62,6 @@
```
AlexanderLanin, line 62: _AI-assisted review (Claude):_

**Important:** This removes a positive test (`#EXPECT`) for link validation against a non-existent parent. The prior EXPECT-NOT tests were no-ops — this EXPECT test is different. Removing it drops the assertion that the rule produces a warning for unknown outgoing links entirely.

The likely root cause: the warning from sphinx-needs (`unknown outgoing link`) doesn't contain `"score_metamodel"`, so it gets filtered out by the new pre-filter on line 208. But the fix should be to handle that at the filter level, not to silently drop test coverage.

src/extensions/score_metamodel/tests/rst/id_contains_feature/test_id_contains_feature.rst
```diff
@@ -13,13 +13,6 @@
    # *******************************************************************************
```
AlexanderLanin, line 13: _AI-assisted review (Claude):_

**Important:** This test verified that "feature deeper in path is explicitly allowed" behavior. If the old EXPECT-NOT was a no-op (as claimed), the rule's behavior for this case was never actually verified. Two outcomes are possible:

- The rule correctly does NOT warn → the test should be kept with a generic form like `#EXPECT-NOT: Feature 'test' not in path`
- The rule incorrectly DOES warn → removing the test hides the regression instead of fixing the rule

Please verify by running the test locally with the fix applied. If the rule is correct, restore this test in the new generic style.

src/extensions/score_metamodel/tests/rst/options/test_options_options.rst
```diff
@@ -61,14 +61,14 @@
    :sufficient: QM
 
 
-#EXPECT-NOT: feat_saf_fmea__test__good_2.sufficient (yes): does not follow pattern `^(yes|no)$`.
+#EXPECT-NOT: pattern
```
AlexanderLanin, line 64: _AI-assisted review (Claude):_

**Suggestion:** `"safety"` and `"pattern"` are very broad substrings. For example, `"safety"` appears in many warning messages in this file — if the test fixture happens to generate an unrelated safety warning at the same line (e.g., from a different check), the EXPECT-NOT will silently pass regardless. A slightly more specific fragment — e.g., `"does not follow pattern"` or `"safety level"` — would remain generic enough to survive message rewording while being far less likely to match false positives.

This is a tradeoff the author can judge, but worth flagging for the safety-critical context of this codebase.

src/extensions/score_metamodel/tests/rst/graph/test_metamodel_graph.rst
```diff
@@ -62,17 +62,6 @@
 
 
 
-.. Parent requirement does not exist
-#EXPECT: unknown outgoing link
```
a-zw, line 66: Why remove this test?

a-zw (approved):

src/extensions/score_metamodel/tests/rst/graph/test_metamodel_graph.rst
MaximilianSoerenPollak: This is a sphinx-needs warning / error. Not one of our own. I do not think we should start testing sphinx-needs functionality.
This did not work before either (as we filter on just errors coming from score_metamodel).

src/extensions/score_metamodel/tests/rst/id_contains_feature/test_id_contains_feature.rst
MaximilianSoerenPollak: I could re-enable this test by renaming the need, or creating more folders.

As the current folder structure is not carried onto the error message (it's just '<folder>/<filename>' it doesn't go deeper than 1 folder). 

What would be preferred here?

src/extensions/score_metamodel/tests/rst/options/test_options_options.rst
MaximilianSoerenPollak: I'm unsure if making it more specific is needed / good. 

In this line we would not expect ANY warning regarding pattern or safety. So I think keeping it broad makes more sense? Thoughts?

src/extensions/score_metamodel/tests/rst/graph/test_metamodel_graph.rst
MaximilianSoerenPollak: This is a sphinx-needs warning / error. Not one of our own. I do not think we should start testing sphinx-needs functionality.
This did not work before either (as we filter on just errors coming from score_metamodel).

src/extensions/score_metamodel/tests/test_rules_file_based.py
MaximilianSoerenPollak: It doesn't narrow the scope silently, it narrows the scope to exactly what we want to test. 

I will double check the documentation to make sure this is mentioned clearly in there.

MaximilianSoerenPollak:
@FScholPer  Just noticed a merge commit on this, was that by accident on this branch?

a-zw (approved):
LGTM

(maybe undo the main-merge for cleanliness)

AlexanderLanin (approved):

PR #539, opened 2026-05-19
Add valid only allowed to link to valid check

MaximilianSoerenPollak:
## 📌 Description
Add valid only allowed to link to valid check

The test is currently only a 'soft warning' and not enabled. (is_new_check =true). 
In future releases this should be enabled and the RST based tests enbaled as well as the metamodel.yaml graph based tests adapted. 

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/extensions/score_metamodel/checks/graph_checks.py
+++ b/src/extensions/score_metamodel/checks/graph_checks.py
@@ -23,7 +23,7 @@ from score_metamodel import (
 from sphinx.application import Sphinx
 from sphinx_needs.config import NeedType
 from sphinx_needs.data import NeedsView
-from sphinx_needs.need_item import NeedItem, NeedLink
+from sphinx_needs.need_item import NeedItem
 
 
 def eval_need_check(need: NeedItem, check: str, log: CheckLogger) -> bool:
--- a/src/extensions/score_metamodel/checks/graph_checks.py
+++ b/src/extensions/score_metamodel/checks/graph_checks.py
@@ -226,5 +226,5 @@ def check_valid_only_links_to_valid(
         )
         invalid_needs = all_linked_needs.difference(valid_needs_id_all)
         if invalid_needs:
-            msg = f"Is valid but links to invalid need(s): {invalid_needs}"
+            msg = f"is valid but links to invalid need(s): {invalid_needs}"
             log.warning_for_need(need, msg, is_new_check=True)
--- a/src/extensions/score_metamodel/checks/graph_checks.py
+++ b/src/extensions/score_metamodel/checks/graph_checks.py
@@ -11,6 +11,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 import operator
+from itertools import chain
 from collections.abc import Callable
 from functools import reduce
 from typing import Any, cast
@@ -22,7 +23,7 @@ from score_metamodel import (
 from sphinx.application import Sphinx
 from sphinx_needs.config import NeedType
 from sphinx_needs.data import NeedsView
-from sphinx_needs.need_item import NeedItem
+from sphinx_needs.need_item import NeedItem, NeedLink
 
 
 def eval_need_check(need: NeedItem, check: str, log: CheckLogger) -> bool:
@@ -198,3 +199,28 @@ def check_metamodel_graph(
                             f" Explanation: {explanation}"
                         )
                         log.warning_for_need(need, msg)
+
+
+@graph_check
+def check_valid_only_links_to_valid(
+    app: Sphinx,
+    all_needs: NeedsView,
+    log: CheckLogger,
+):
+    # Get all possible link types
+    needs_dict_all = {need["id"]: need for need in all_needs.values()}
+    needs_dict_local = {
+        need["id"]: need for need in all_needs.filter_is_external(False).values()
+    }
+    # Pre-Filter for only valid & local needs
+    valid_needs_id_all = [x.id for x in all_needs.values() if x.get("status") == "valid"]
+    valid_needs_local = [
+        x for x in all_needs.filter_is_external(False).values() if x.get("status") == "valid"
+    ]
+
+    for need in valid_needs_local:
+        all_linked_needs: list[NeedLink] = list(chain(*need._links.values()))  # type: ignore
+        for link in all_linked_needs:
+            if link.id not in valid_needs_id_all:
+                msg = f"Is valid but links to invalid need: {link.id}"
+                log.warning_for_need(need, msg, is_new_check=True)
--- a/src/extensions/score_metamodel/checks/graph_checks.py
+++ b/src/extensions/score_metamodel/checks/graph_checks.py
@@ -207,11 +207,11 @@ def check_valid_only_links_to_valid(
     all_needs: NeedsView,
     log: CheckLogger,
 ):
-    # Get all possible link types
-    # Pre-Filter for only valid & local needs
-    valid_needs_id_all = [
+    # Pre-Gather all *valid* need id's (external, & local)
+    valid_needs_id_all = set(
         x.id for x in all_needs.values() if x.get("status") == "valid"
-    ]
+    )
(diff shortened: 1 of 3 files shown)


src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -198,3 +199,28 @@ def check_metamodel_graph(
...
+    log: CheckLogger,
+):
+    # Get all possible link types
+    # Pre-Filter for only valid & local needs
+    valid_needs_id_all = [
+        x.id for x in all_needs.values() if x.get("status") == "valid"
+    ]
+    valid_needs_local = [
+        x
+        for x in all_needs.filter_is_external(False).values()
```
a-zw, line 217: Why filter out externals? Doesn't that mean it is ok to link invalid external needs?

src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -198,3 +199,28 @@ def check_metamodel_graph(
...
+        if x.get("status") == "valid"
+    ]
+
+    for need in valid_needs_local:
+        all_linked_needs: list[NeedLink] = list(chain(*need._links.values()))  # type: ignore
+        for link in all_linked_needs:
+            if link.id not in valid_needs_id_all:
```
a-zw: I'm a little worried about performance here because we will eventually have to process many needs items. Using lists this is O(n³). Just by using a set for `valid_needs_id_all`, this should be O(n²) which is bad enough.

src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -198,3 +199,28 @@ def check_metamodel_graph(
...
+    ]
+    valid_needs_local = [
+        x
+        for x in all_needs.filter_is_external(False).values()
+        if x.get("status") == "valid"
+    ]
+
+    for need in valid_needs_local:
+        all_linked_needs: list[NeedLink] = list(chain(*need._links.values()))  # type: ignore
+        for link in all_linked_needs:
```
a-zw: Can we check `link.get("status") == "valid"` instead of checking if the id is in some list?

a-zw (changes requested):

src/extensions/score_metamodel/checks/graph_checks.py
MaximilianSoerenPollak: If you look at the check, what we want to do is we want to check that any LOCAL needs do not link to any invalid needs (local or external). 

This way we will not check that external ones link badly.

src/extensions/score_metamodel/checks/graph_checks.py
MaximilianSoerenPollak: Not quiet sure I can follow the difference here. 
Where would having a set instead of a list make a difference in this? 

Yes lists isn't ideal already you are right, that's why I used the itertools chain to ensure we have the fastest way to group the links.

src/extensions/score_metamodel/checks/graph_checks.py
MaximilianSoerenPollak: We could theoretically, but the issue is that the link we get here (NeedLink) is **not** a real Need. 
We would then have to use that NeedLink.id to look up the actuall need in a dictionary. 

So I thought it will be faster to gather all the 'valid needs' once, and then just check the id.

src/extensions/score_metamodel/checks/graph_checks.py
a-zw: With a list the `not in` check works by iterating over the list O(n). With a set, it is a hash-map-lookup O(1).

PandaeDo (dismissed):
LGTM

AlexanderLanin (commented):
*AI-assisted review (Claude):*

**Critical: 0 | Important: 2 | Suggestions: 0**

The check logic is sensible, but there are two issues that need attention before merge.

**1. Test file has no assertions (Important)**
`test_invalid_graph.rst` builds successfully but verifies nothing. The `#EXPECT:` line is inside an RST comment (`..`), so `test_rules_file_based.py` never parses it — the test framework matches `line.startswith("#EXPECT:")`, which is false for RST-commented lines. If the check function were deleted tomorrow, this test would still pass. The PR description calls this intentional (soft warning / `is_new_check=True`), but the result is a test file with zero test value. Consider either removing it until the check is promoted to a real warning, or adding a mechanism to verify INFO-level check output.

On top of this, the commented-out expected message references `comp_saf_fmea__child__16`, but the defined need has ID `comp_saf_fmea__child__1` — so even if the line were uncommented, it would not match.

**2. Dead links and invalid-status links conflated (Important)**
If a valid need links to a non-existent ID (dead link), `link.id not in valid_needs_id_all` is also true, producing `"is valid but links to invalid need: <dead-id>"` when the linked need simply doesn't exist. This produces a misleading message and potentially duplicates what sphinx-needs already reports for dead links. The check should distinguish between a missing need and one with the wrong status.

**On existing comments:**
- The `list` vs `set` performance point raised by @a-zw is correct — `valid_needs_id_all` should be a `set` for O(1) lookup. An even cleaner fix is to drop the pre-collection entirely and resolve each `link.id` directly via `all_needs.get(link.id)`, which also handles the dead-link case above and avoids the `_links` private attribute access.

src/extensions/score_metamodel/checks/graph_checks.py
MaximilianSoerenPollak: Ohh, I did not know. I will adapt that for sure. I thought sets and lists are same lookup.

src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -198,3 +199,28 @@ def check_metamodel_graph(
...
+def check_valid_only_links_to_valid(
+    app: Sphinx,
+    all_needs: NeedsView,
+    log: CheckLogger,
+):
+    # Get all possible link types
+    # Pre-Filter for only valid & local needs
+    valid_needs_id_all = [
+        x.id for x in all_needs.values() if x.get("status") == "valid"
+    ]
```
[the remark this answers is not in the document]
a-zw: Did it read my comment? 🤣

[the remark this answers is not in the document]
MaximilianSoerenPollak: Stolen valor :D

src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -198,3 +199,28 @@ def check_metamodel_graph(
...
+    valid_needs_local = [
+        x
+        for x in all_needs.filter_is_external(False).values()
+        if x.get("status") == "valid"
+    ]
+
+    for need in valid_needs_local:
+        all_linked_needs: list[NeedLink] = list(chain(*need._links.values()))  # type: ignore
```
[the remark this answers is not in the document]
MaximilianSoerenPollak: There is no way around this that is better. 
We already have all of the links inside the need element, I can access it like this easily. 

Going around and doing it via the app.get_all_link_rypes and then CHECKING each link type if it's possible / doable in this need and valid, just seems very workaroundy and also much slower.

src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -198,3 +199,32 @@ def check_metamodel_graph(
...
+
+    for need in valid_needs_local:
+        # Using set comprehension here to enable faster computation for comparisons
+        all_linked_needs: set[str] = set(
+            x.id
+            for x in set(chain(*need._links.values()))  # type: ignore
+        )
+        invalid_needs = all_linked_needs.difference(valid_needs_id_all)
+        if invalid_needs:
+            msg = f"Is valid but links to invalid need(s): {invalid_needs}"
```
a-zw:
```suggestion
            msg = f"is valid but links to invalid need(s): {invalid_needs}"
```

So it matches the disabled test.

a-zw (approved):

AlexanderLanin (approved):

PR #541, opened 2026-05-19
PoC DR-008 Option B

a-zw:
Just a proof-of-concept!

A real implementation should consider backwards-compatibility, for example.


a-zw:
Closing as we picked option M instead.

PR #544, opened 2026-05-20
Feat: Forbid certain uses of need extends directive

### LINKED ISSUE #345: Docs as code allow to implies changes which causes side effects in other modules/features/components
Docs as code provide the needextend function from sphinx. But the sphinx needs objects are a kind of global variables. 
If I extend or change a needs object in a local module (reviewed in a pull request) then it might implicate changes in any other part which uses this sphinxs object.

For example: If I change locally the ASIL level of an component from ASIL to QM, because I only use it as QM, that might be complete fine for my part and therefore nobody will comply in the review of that change. But in any other part which are not changed by my PR the classification of this sphinx needs object changes as well. On the PR will not show where this change happens.

MaximilianSoerenPollak:
## 📌 Description
Adds capability to check needextend usage and throw warnings if used in a way we do not allow

Fixes: #345 

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs/how-to/write_docs.rst
+++ b/docs/how-to/write_docs.rst
@@ -15,6 +15,9 @@
 Write Documentation
 ===================
 
+Outside Documentation
+^^^^^^^^^^^^^^^^^^^^^
+
 `Sphinx <https://www.sphinx-doc.org/en/master/>`_:
 the documentation generator we use.
 
@@ -23,3 +26,46 @@ the plain-text markup language used for most source files in this project.
 
 `Sphinx-Needs <https://sphinx-needs.readthedocs.io/en/latest/>`_:
 a Sphinx extension that models requirements, tests, tasks and other "needs" inside the docs.
+
+
+Advanced
+^^^^^^^^
+
+Needextend
+~~~~~~~~~~
+
+Needextend allows you to extend needs that are defined in the documentaiton.
+The scope of allowed behaviour in Docs-As-Code for needextend is limited as we do not allow all of its usecases.
+
+Allowed uscases: 
+- Setting an attribute or Link **IF** it is not already set in the need that is getting modified
+- Appending to a list of links
+
+Not Allowed:
+- Overwriting an attribute or Link that already is set in the need that gets modified
+- Deleting an attribute or Link
+
+.. code-block:: rst
+
+      .. stkh_req:: Test Req Extends 1
+         :id: stkh_req__test__need_extends_1
+         :status: invalid
+
+
+      # ✅ ALLOWED => The replacing of attributes on needs that are NOT set.
+      .. needextend:: c.this_doc() and id == 'stkh_req__test__need_extends_1'
+         :safety: NO 
+
+
+      # ❌ NOT ALLOWED => Overwriting attributes that are already set in the need
+      .. needextend:: c.this_doc() and id == 'stkh_req__test__need_extends_1'
+         :status: valid
+
+
+For further documentation on needextends please `look here <https://sphinx-needs.readthedocs.io/en/latest/directives/needextend.html>`_
+
+.. note:: 
+
+   In the future we will enable a check that needextends will only modify needs in the current document. 
+   You can ensure this by adding `c.this_doc()` to the filter string of the need.
+
--- a/docs/how-to/write_docs.rst
+++ b/docs/how-to/write_docs.rst
@@ -38,12 +38,14 @@ Needextend allows you to extend needs that are defined in the documentaiton.
 The scope of allowed behaviour in Docs-As-Code for needextend is limited as we do not allow all of its usecases.
 
 Allowed uscases:
-- Setting an attribute or Link **IF** it is not already set in the need that is getting modified
-- Appending to a list of links
+
+* Setting an attribute or Link **IF** it is not already set in the need that is getting modified
+* Appending to a list of links
 
 Not Allowed:
-- Overwriting an attribute or Link that already is set in the need that gets modified
-- Deleting an attribute or Link
+
+* Overwriting an attribute or Link that already is set in the need that gets modified
+* Deleting an attribute or Link
 
 .. code-block:: rst
 
--- a/docs/how-to/write_docs.rst
+++ b/docs/how-to/write_docs.rst
@@ -34,10 +34,10 @@ Advanced
 Needextend
 ~~~~~~~~~~
 
-Needextend allows you to extend needs that are defined in the documentaiton.
+Needextend allows you to extend needs that are defined in the documentation.
 The scope of allowed behaviour in Docs-As-Code for needextend is limited as we do not allow all of its usecases.
 
-Allowed uscases:
+Allowed usecases:
 
 * Setting an attribute or Link **IF** it is not already set in the need that is getting modified
 * Appending to a list of links
--- a/docs/how-to/write_docs.rst
+++ b/docs/how-to/write_docs.rst
@@ -37,7 +37,7 @@ Needextend
 Needextend allows you to extend needs that are defined in the documentaiton.
 The scope of allowed behaviour in Docs-As-Code for needextend is limited as we do not allow all of its usecases.
 
-Allowed uscases: 
+Allowed uscases:
 - Setting an attribute or Link **IF** it is not already set in the need that is getting modified
 - Appending to a list of links
 
@@ -54,7 +54,7 @@ Not Allowed:
 
       # ✅ ALLOWED => The replacing of attributes on needs that are NOT set.
       .. needextend:: c.this_doc() and id == 'stkh_req__test__need_extends_1'
-         :safety: NO 
+         :safety: NO
 
 
       # ❌ NOT ALLOWED => Overwriting attributes that are already set in the need
@@ -64,8 +64,7 @@ Not Allowed:
 
 For further documentation on needextends please `look here <https://sphinx-needs.readthedocs.io/en/latest/directives/needextend.html>`_
 
-.. note:: 
+.. note::
 
-   In the future we will enable a check that needextends will only modify needs in the current document. 
+   In the future we will enable a check that needextends will only modify needs in the current document.
    You can ensure this by adding `c.this_doc()` to the filter string of the need.
-
--- a/src/extensions/score_metamodel/checks/check_needs_extends.py
+++ b/src/extensions/score_metamodel/checks/check_needs_extends.py
@@ -81,7 +81,7 @@ def score_extend_needs_data_func(  # noqa: C901
             if found_need["is_external"]:
                 log_warning(
                     logger,
-                    f"Error when extending need: {found_need['id']} "
+                    f"Error when extending need: {found_need['id']}. "
                     + "It is not allowed to modify external needs via needextend",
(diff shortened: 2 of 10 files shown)


src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -53,6 +53,12 @@ prohibited_words_checks:
       - thing
       - absolutely
 
+unmutable_options:
```
a-zw: The word is "immutable".

AlexanderLanin (commented):

**Critical: 0 | Important: 2 | Suggestions: 3**

---

### Overall

Solid addition — the monkey-patch target is correct (`sphinx_needs.directives.need` imports and calls `extend_needs_data` locally, so patching that module's namespace intercepts the call). The approach of wrapping the upstream function and delegating is clean.

Two concerns worth addressing before merge:

1. **The check cannot be disabled via `score_metamodel_checks`** — it fires unconditionally at import time rather than integrating with the existing check registry.
2. **Three of the five validation branches have no test coverage** — link REPLACE/DELETE, field DELETE, and APPEND on string options are all untested.

See inline comments for details.

src/extensions/score_metamodel/checks/check_needs_extends.py
```diff
@@ -0,0 +1,134 @@
...
+                        if need[option_name]:
+                            error_msg = f"Error when extending need: {need['id']}. Replacing of options that are already set is not allowed via needextends."
+
+                            log_warning(
+                                logger, error_msg, "needextend", location=location
+                            )
+    return original_function(all_needs, extends, needs_config)
+
+
+sphinx_needs.directives.need.extend_needs_data = score_extend_needs_data_func
```
AlexanderLanin, line 142: _AI-assisted review (Claude):_

**Important: Check is hardwired at import time, bypassing the registry**

The last line patches the function unconditionally when the module is imported via `discover_checks()`. This means the check:

- always runs, even when `score_metamodel_checks` is set to a filter list that excludes it
- cannot be tested in isolation with `#CHECK:` annotations
- cannot be turned off for a specific repo via config

All other checks in `checks/` use `@local_check` or `@graph_check` to self-register. This one side-steps that entirely.

The issue is structural: the validation logic here runs inside the `extend_needs_data` call itself (which happens during the Sphinx build phase, not in `_run_checks`), so it can't directly use the `CheckLogger` / registry pattern without a bigger refactor. But at a minimum the check should respect the filter — either read `app.config.score_metamodel_checks` at call time or document explicitly why this one is always-on.

src/extensions/score_metamodel/tests/rst/options/test_need_extends.rst
```diff
@@ -0,0 +1,42 @@
...
+
+.. needextend:: c.this_doc() and id == 'stkh_req__test__need_extends_1'
+   :safety: NO
+
+
+.. This will be activated once we have activated the c.this_doc() check aswell
+.. #EXPECT: Potentially altering needs outside of the document is not allowed. Please add 'c.this_doc()' to the needextend to limit it to only needs in the same document
+
+.. .. needextend:: id == 'stkh_req__test__need_extends_1'
+.. :security: QM
```
AlexanderLanin, line 75: _AI-assisted review (Claude):_

**Important: Three validation branches have no test coverage**

The checker has five distinct warning paths. Only two are exercised here:

| Branch | Covered? |
|--------|----------|
| `list_modifications`: REPLACE/DELETE on links | ❌ |
| `modifications`: DELETE | ❌ |
| `modifications`: APPEND on string option | ❌ |
| `modifications`: REPLACE when option already set | ✅ (line 23) |
| `modifications`: REPLACE when option **not** set | ✅ (line 32) |

The three uncovered branches represent ~60% of the warning logic. Adding RST fixtures for at least `list_modifications` replace-on-links and `DELETE` would significantly increase confidence that the logic works as intended.

src/extensions/score_metamodel/checks/check_needs_extends.py
```diff
@@ -0,0 +1,134 @@
...
+                )
+                continue
+        for found_need in found_needs:
+            if found_need["is_external"]:
+                log_warning(
+                    logger,
+                    "Error. It is not allowed to modify external needs via needextend",
+                    "needextend",
+                    location,
+                )
```
AlexanderLanin, line 88: _AI-assisted review (Claude):_

**Suggestion: external-need branch warns but continues processing**

After logging the warning for an external need, the loop falls through and still executes the modification checks below (lines 89–130) for that same need. The original `extend_needs_data` will then apply the actual change. This means the user gets a warning *and* the modification still happens.

If modifying external needs is genuinely disallowed, this should `continue` after the warning (or the original function should skip it too). If it's just advisory for now, the message "It is not allowed" is misleading — consider "Warning: modifying external needs is discouraged" until it's enforced.

src/extensions/score_metamodel/checks/check_needs_extends.py
```diff
@@ -0,0 +1,134 @@
...
+            # Work in the stored needs, not on the search result
+            need = all_needs[found_need["id"]]
+
+            location = (
+                current_needextend["docname"],
+                current_needextend["lineno"],
+            )
```
AlexanderLanin, line 95: _AI-assisted review (Claude):_

**Suggestion: `location` is re-assigned inside the per-need loop, shadowing the outer variable**

`location` is first set on line 44 for the current `needextend` directive, then reassigned on lines 91–94 inside the `for found_need in found_needs` loop. The second assignment is identical to the first — it just copies the same values. This creates a subtle trap: if a future change needs the needextend's location before the inner loop, the pre-loop assignment on line 44 looks authoritative but gets clobbered. Remove the redundant inner reassignment.

docs/how-to/write_docs.rst
```diff
@@ -23,3 +26,47 @@ the plain-text markup language used for most source files in this project.
...
 a Sphinx extension that models requirements, tests, tasks and other "needs" inside the docs.
+
+
+Advanced
+^^^^^^^^
+
+Needextend
+~~~~~~~~~~
+
+Needextend allows you to extend needs that are defined in the documentaiton.
```
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion: typos in the new documentation**

Line 37: `documentaiton` → `documentation`
Line 40: `Allowed uscases` → `Allowed usecases`

src/extensions/score_metamodel/checks/check_needs_extends.py
MaximilianSoerenPollak: How are we suppose to add this to local_check or graph_check mr Robot. 

This is a wrapper, that HAS to take the arguments that the function takes, we can not change that. 
This is not viable.

src/extensions/score_metamodel/checks/check_needs_extends.py
MaximilianSoerenPollak: Ye that's not an issue, you get the warning and the build fails. 
Though we could do early exits to speed it up, that mightn ot be a bad idea.

src/extensions/score_metamodel/checks/check_needs_extends.py
MaximilianSoerenPollak: That is taken stragith from the function. I'm not changing that, it is how the original function works.

AlexanderLanin (approved):

src/extensions/score_metamodel/tests/rst/options/test_need_extends.rst
```diff
@@ -0,0 +1,75 @@
...
+   :safety: NO
+
+
+#EXPECT-NOT: Replacing of options
+
+.. needextend:: c.this_doc() and id == 'stkh_req__test__need_extends_1'
+   :safety: NO
+
+
+# EXPECT: Replace or Delete action is not allowed via needextends.
```
AlexanderLanin: _AI-assisted review (Claude):_

**Critical: `# EXPECT:` (space) is silently ignored by the test parser**

Line 53 uses `# EXPECT:` with a space after `#`. The parser in `test_rules_file_based.py` checks `line.startswith("#EXPECT:")` — the space breaks the match, so this annotation is parsed as an ordinary RST comment and discarded.

Result: the `list_modifications` REPLACE/DELETE branch (`check_needs_extends.py` lines 96–107) has **no effective test**. A bug in that branch would not be caught.

Fix: remove the space.

```suggestion
#EXPECT: Replace or Delete action is not allowed via needextends.
```

AlexanderLanin (commented):

**Critical: 1 | Important: 0 | Suggestions: 0**

---

### New finding

One new critical issue found in the latest commits: **line 53 of `test_need_extends.rst` uses `# EXPECT:` (with a space) instead of `#EXPECT:`**. The parser uses `line.startswith("#EXPECT:")` so the space causes the annotation to be silently dropped. The REPLACE/DELETE links branch in `check_needs_extends.py` (lines 96–107) therefore has no effective test. See inline comment for a one-line suggestion fix.

---

### Responses to previous review pushback

**Registry integration (my previous Important finding):** The author's response — "this is a wrapper, it has to take those arguments" — misses the point. The concern isn't the function signature; it's that the check fires unconditionally regardless of `score_metamodel_checks`. A minimal fix could be: read `app.config.score_metamodel_checks` at call time inside `score_extend_needs_data_func` and return early if this check is excluded. No signature change needed. That said, if the project is OK with this check being permanently always-on and non-filterable, that's a policy decision — but it should be explicit, not accidental.

**External-need fall-through (my previous Suggestion):** The author says "you get the warning and the build fails." If the build already fails, the fall-through is harmless — but the word "Error" in the warning message combined with processing continuing is still confusing to readers of the log. A `continue` or a message reword is still worth the line.

**Redundant `location` reassignment (my previous Suggestion):** The author says it's taken straight from the original function. That's fine as a reason to not change it — accepted.

---

### Unchanged issues

**Typos** in `docs/how-to/write_docs.rst` (lines 37–40: `documentaiton`, `uscases`) were not fixed. Still worth a one-liner before merge.

AlexanderLanin:
`WARNING: Error. It is not allowed to modify external needs via needextend` should print which need was attempted to modify, that will help improving the filter. Actually this should include a hint to use c.this_doc ?!

AlexanderLanin (changes requested):

AlexanderLanin (commented):

---

The two new commits resolve the critical issue and the test coverage gap:

- The `# EXPECT:` space bug is fixed — all five branches now use correct `#EXPECT:` annotations.
- All five warning paths have explicit test coverage with correct expected strings (verified that EXPECT substrings match the produced error messages).
- Error messages now consistently include the need ID, which is an improvement over the previous generic messages.

One thing still open from the prior round: **doc typos in `docs/how-to/write_docs.rst`** — line 37 (`documentaiton`) and line 40 (`Allowed uscases:`). These have been raised twice now and are a one-line fix.

Otherwise this looks ready.

AlexanderLanin:
_AI fix-commit (Claude):_ `7ae33374` — fix typos: `documentaiton` → `documentation`, `uscases` → `usecases`

AlexanderLanin (approved):

PR #545, opened 2026-05-20
Tracability improvements - needs type autodiscovery

FScholPer:
## 📌 Description
Improved metrics.json by
- autodiscovery

## 🚨 Impact Analysis

- [X] This change does not violate any tool requirements and is covered by existing tool requirements
- [X] This change does not violate any design decisions

## ✅ Checklist

- [X] Added/updated documentation for new or changed features
- [X] Added/updated tests to cover the changes
- [X] Followed project coding standards and guidelines


MaximilianSoerenPollak:
Bit confused about this, can you explain what this is suppose to solve / help with? 
Maybe we encountered this problem already somewhere else and can help with finding a good solution.

FScholPer:
sure the problem occurred in the score repo after adding the new metrics.json from docs-as-code. There was no config set up https://github.com/eclipse-score/docs-as-code/blob/main/docs/how-to/dashboards_and_quality_gates.rst it produced metrics like tool_req which does not make sense in platform repo. I added autodiscovery for requirement types so that explicit definition is not needed as default

MaximilianSoerenPollak:
> sure the problem occurred in the score repo after adding the new metrics.json from docs-as-code. There was no config set up https://github.com/eclipse-score/docs-as-code/blob/main/docs/how-to/dashboards_and_quality_gates.rst it produced metrics like tool_req which does not make sense in platform repo. I added autodiscovery for requirement types so that explicit definition is not needed as default

Okay, still a bit confused though. 
As sphinx-needs gives you automatically all needs types that the current project has registered. 
You just have to look into `app.config.needs_types` and you will get a list with all of them. 
No need to look into any other configuration. Only thing you have to make sure is that you check it after `metamodel/__init__.py line 307` where this actually gets extended.

Would that not be an easier way? Though I might still not 100% understand/grasp it and this is not possible for this.

AlexanderLanin:
we want only requirement types AFAIK

MaximilianSoerenPollak:
> we want only requirement types AFAIK

Yes we chated in slack and I got it explained a bit better and understand now what is wanted here.

FScholPer:
> > sure the problem occurred in the score repo after adding the new metrics.json from docs-as-code. There was no config set up https://github.com/eclipse-score/docs-as-code/blob/main/docs/how-to/dashboards_and_quality_gates.rst it produced metrics like tool_req which does not make sense in platform repo. I added autodiscovery for requirement types so that explicit definition is not needed as default
> 
> Okay, still a bit confused though. As sphinx-needs gives you automatically all needs types that the current project has registered. You just have to look into `app.config.needs_types` and you will get a list with all of them. No need to look into any other configuration. Only thing you have to make sure is that you check it after `metamodel/__init__.py line 307` where this actually gets extended.
> 
> Would that not be an easier way? Though I might still not 100% understand/grasp it and this is not possible for this.

Good point, and agreed that app.config.needs_types is the right source after metamodel setup. In metrics generation we intentionally do one extra step: we filter those registered requirement types to the ones actually present in the current repo needs (respecting include_external). If we only take app.config.needs_types as-is, we can still produce irrelevant entries like tool_req in repos that don’t use it. So the implementation is ‘needs_types-based autodiscovery + present-in-build filtering’.

src/extensions/score_metamodel/__init__.py
```diff
@@ -147,6 +156,61 @@ def _write_metrics_json(app: Sphinx, exception: Exception | None) -> None:
     logger.info(f"Traceability metrics written to: {out_path}")
 
 
+def _get_need_value(need: Any, key: str, default: Any = None) -> Any:
```
a-zw: This seems weird. Sometimes it is a callable and sometimes not and we don't know in advance?

docs/conf.py
```diff
@@ -18,3 +18,7 @@
 extensions = [
     "score_sphinx_bundle",
 ]
+
+# Configure traceability metrics explicitly for this repository.
+score_metamodel_requirement_types = "tool_req"
```
a-zw: If we have autodiscovery, why have this option too?

docs/conf.py
```diff
@@ -18,3 +18,7 @@
...
+score_metamodel_include_external_needs = False
```
a-zw: False is the default, why add this noise here?

src/extensions/score_metamodel/__init__.py
```diff
@@ -147,6 +156,61 @@ def _write_metrics_json(app: Sphinx, exception: Exception | None) -> None:
...
+    present_types: set[str] = set()
+    for need in all_needs:
+        is_external = bool(_get_need_value(need, "is_external", False))
+        if not include_external and is_external:
+            continue
+        need_type: Any = _get_need_value(need, "type", None)
+        if isinstance(need_type, str):
+            present_types.add(need_type)
+    discovered = tagged_requirements.intersection(present_types)
+    if not discovered:
```
a-zw: Why do we need this fallback?

src/extensions/score_metamodel/__init__.py
```diff
@@ -147,6 +156,61 @@ def _write_metrics_json(app: Sphinx, exception: Exception | None) -> None:
...
+        if directive:
+            tagged_requirements.add(directive)
+
+    present_types: set[str] = set()
+    for need in all_needs:
+        is_external = bool(_get_need_value(need, "is_external", False))
+        if not include_external and is_external:
+            continue
+        need_type: Any = _get_need_value(need, "type", None)
+        if isinstance(need_type, str):
```
a-zw: Is this check necessary? Maybe we rather want it to throw here to figure out what went wrong instead of silently ignoring some types?

a-zw (changes requested):

src/extensions/score_metamodel/__init__.py
AlexanderLanin: _AI-assisted review (Claude):_

**Important — `_req` suffix fallback undermines the tag-based approach**

The primary discovery logic (tag intersection) is clean and well-motivated. This fallback, however, is a name-based heuristic that can silently produce wrong results — any type whose directive happens to end in `_req` gets treated as a requirement.

In the current metamodel this is accidentally safe because all `*_req` directives are in fact requirements. But the fallback makes the system depend on a naming convention rather than the metamodel tags, which is the opposite of what autodiscovery was designed to achieve.

If no tagged requirement types match what's in the build, that's a signal that the repo genuinely has no requirement needs — returning an empty set (and writing empty metrics) is the correct behavior. If it's a misconfiguration, a warning is more helpful than silently guessing.

Suggestion: drop the fallback entirely. If you want to preserve a safety net, log a warning when `tagged_requirements` is non-empty but `discovered` is empty (tags exist in metamodel but none are present in the build).

This aligns with @a-zw's comment on this line.

src/extensions/score_metamodel/__init__.py
AlexanderLanin: _AI-assisted review (Claude):_

**Important — over-defensive access pattern with bare `except Exception`**

In sphinx-needs, `NeedItem` is a `TypedDict` (dict subclass). The `get` method always works, and `need["type"]` always works if the key exists. The dual-access pattern (check for callable `.get`, fall back to `__getitem__`, swallow all exceptions) obscures bugs — if a need is missing `"type"` or `"is_external"`, that is a data integrity problem that should surface, not be silently defaulted away.

The test `_FakeObjectNeedsData` / `_NeedObj` exercises this code path, but it tests a scenario that does not occur in production. In practice, `SphinxNeedsData.get_needs_view()` returns `NeedsView` (a dict of `NeedItem` dicts).

Suggestion: use plain `need.get(key, default)` or `need[key]`. If future sphinx-needs versions change the return type, that will be caught by type checking and test failures rather than silently falling through to a default.

This aligns with @a-zw's comment on this function.

docs/conf.py
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion — explicit config in this repo contradicts the autodiscovery purpose**

Setting `score_metamodel_requirement_types = "tool_req"` here hard-codes the exact behavior that autodiscovery was designed to replace. This repo only has `tool_req` needs, so autodiscovery would produce the same result automatically.

`score_metamodel_include_external_needs = False` is the default — adding it is noise, as @a-zw noted.

Recommend removing both lines. The docs-as-code repo should dogfood autodiscovery to validate it works.

AlexanderLanin (commented):

**Critical: 0, Important: 2, Suggestions: 2**

The autodiscovery approach is sound — intersecting metamodel-tagged requirement types with types actually present in the build is the right way to avoid irrelevant metrics entries like `tool_req` in repos that don't use it.

Two structural concerns with the current implementation, plus two minor items.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -50,6 +50,11 @@ In ``docs/conf.py``:
    score_metamodel_requirement_types = "feat_req,comp_req,aou_req"
    score_metamodel_include_external_needs = False
 
+By default, ``score_metamodel`` autodiscovers requirement types from the
+repository needs that are present in the current build. If
+``score_metamodel_requirement_types`` is set, that explicit list overrides
+autodiscovery.
+
```
AlexanderLanin, line 67: _AI-assisted review (Claude):_

**Suggestion — documentation should lead with autodiscovery, not the override**

The "Minimal Configuration Example" still shows the explicit `score_metamodel_requirement_types` setting as the primary pattern, with autodiscovery mentioned as an afterthought below. Since the whole point of this PR is that explicit config should be unnecessary by default, consider restructuring:

1. State that autodiscovery is the default (no config needed).
2. Then show the explicit override as an escape hatch for repos that need it.

This better communicates the intended user experience.

FScholPer:
Updated with fixes

src/extensions/score_metamodel/__init__.py
```diff
@@ -147,6 +156,59 @@ def _write_metrics_json(app: Sphinx, exception: Exception | None) -> None:
     logger.info(f"Traceability metrics written to: {out_path}")
 
 
+def _get_need_value(need: Any, key: str, default: Any = None) -> Any:
+    return need.get(key, default)
```
MaximilianSoerenPollak: This abstraction makes no sense. 
You use this once.  Abstraction just for the sake of abstraction is not good.

src/extensions/score_metamodel/__init__.py
```diff
@@ -147,6 +156,59 @@ def _write_metrics_json(app: Sphinx, exception: Exception | None) -> None:
...
+    if not isinstance(need_type, dict):
+        return None
+    directive = need_type.get("directive")
+    tags = need_type.get("tags", [])
+    if not isinstance(directive, str) or not isinstance(tags, list):
+        return None
+    normalized = {str(tag).strip() for tag in tags}
+    if "requirement_excl_process" in normalized or "requirement" in normalized:
+        return directive
+    return None
```
MaximilianSoerenPollak: I don't quiet understand this. 

What exactly do you want to get here, all need types that are directives? 
That would be all types you get from `app.config.need_types`. 

If you want to get our own `ScoreNeedType` which defines some additional things which makes checking easier, see signature: 
```python
class ScoreNeedType(NeedType):
    tags: list[str]
    parts: int

    mandatory_options: dict[str, str]
    optional_options: dict[str, str]

    mandatory_links: AllowedLinksType
    optional_links: AllowedLinksType
```
Then you could just copy the function we have in `check_options` which goes like:
```python
def get_need_type(needs_types: list[ScoreNeedType], directive: str) -> ScoreNeedType:
    for need_type in needs_types:
        assert isinstance(need_type, dict), need_type
        if need_type["directive"] == directive:
            return need_type
    raise ValueError(f"Need type {directive} not found in needs_types")
```
And adapt it so you I guess can filter directly for the tag as well so you don't waste itterations: 
```python
def get_need_type(needs_types: list[ScoreNeedType], directive: str) -> ScoreNeedType:
    for need_type in needs_types:
        assert isinstance(need_type, dict), need_type
        if need_type["directive"] == directive and ("requirement_excl_process" in need_type["tags"] or "requirement" in need_type["tags"]:
            return need_type
    raise ValueError(f"Need type {directive} not found in needs_types")
```

src/extensions/score_metamodel/__init__.py
```diff
@@ -147,6 +156,59 @@ def _write_metrics_json(app: Sphinx, exception: Exception | None) -> None:
...
+def _discover_requirement_types(
+    app: Sphinx, all_needs: list[Any], include_external: bool
+) -> set[str]:
+    """Discover requirement directives that are both tagged and present."""
+    tagged_requirements: set[str] = set()
+    needs_types = getattr(app.config, "needs_types", [])
+    for need_type in needs_types or []:
+        directive = _as_requirement_directive(need_type)
+        if directive:
+            tagged_requirements.add(directive)
```
MaximilianSoerenPollak: app.config.need_types will give you all need types that are possible. 
If you then iterate over all needs you will **only** get present needs. Sphinx-needs will not give you empty needs that aren''t defined in the project. 
So in my opinion this filter makes not much sense.

You could get all needs like this: 
```
all_local_needs = SphinxNeedsData(app.env).get_needs_view().filter_is_external(False)
```
This already filters the needs and exludes all external ones.

MaximilianSoerenPollak (commented):
Some more questions.

FScholPer:
Thanks for the comments will continue next week...

src/extensions/score_metrics/traceability_metrics.py
```diff
@@ -0,0 +1,195 @@
...
+    override_path = Path(raw_metamodel_path) if raw_metamodel_path else None
+    metamodel = load_metamodel_data(override_path)
+
+    # We either get the types that should be considered from the configuration
+    # If none are specified the need types declared in the Metamodel with
+    # the tags 'requirement' are taken
+    raw = getattr(app.config, "score_metamodel_requirement_types", "").strip()
+    filter_reqs = [t.strip() for t in raw.split(",") if t.strip()]
+    if not filter_reqs:
+        filter_reqs = get_need_types_by_tags(metamodel.needs_types, {"requirement"})
```
AlexanderLanin, line 141: _AI-assisted review (Claude):_

**Important — autodiscovery tag set contradicts the config description**

`score_metrics/__init__.py` documents that autodiscovery uses tags `"requirement, requirement_excl_process"`, but the actual call in `traceability_metrics.py:141` only passes `{"requirement"}`:

```python
filter_reqs = get_need_types_by_tags(metamodel.needs_types, {"requirement"})
```

Any type tagged *only* with `requirement_excl_process` (without also carrying `requirement`) will not be autodiscovered, silently producing incorrect metrics. Either:

1. Fix the call to pass `{"requirement", "requirement_excl_process"}`, or
2. Fix the config description to match what the code actually does.

The right answer depends on whether `requirement_excl_process` types should appear in `metrics.json`. That is a design decision — but right now the doc and the code disagree.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -136,6 +151,296 @@ For a new consumer repository:
 4. Introduce modest thresholds in CI.
 5. Raise thresholds over time as the repository matures.
 
+..
+ ╓                                                           ╖
+ ║ .. The Following part has been generated by Copilot       ║
```
AlexanderLanin: _AI-assisted review (Claude):_

**Important — AI-generated Copilot content committed into user-facing documentation**

Line 156 marks a ~295-line section of `dashboards_and_quality_gates.rst` as AI-generated by Copilot. This section documents internal Python functions (`_get_key_values`, `get_metrics_with_overall_total_considered`, etc.) in exhaustive API-reference style — including testing recommendations, common pitfalls, and notes about `print()` side effects — inside a user-facing how-to guide.

This content does not belong in user-facing docs and should not have been committed as-is. Please remove or replace it with curated documentation.

src/extensions/score_metrics/sphinx_filters.py
```diff
@@ -39,6 +39,9 @@ def func(needs: list[NeedItem], results: list[int], **kwargs) -> None: ...
 
 from __future__ import annotations
 
+from typing import Any
+
+from score_metrics.traceability_metrics import CALCULATED_METRICS
 from sphinx_needs.need_item import NeedItem
```
AlexanderLanin, line 45: _AI-assisted review (Claude):_

**Suggestion — stale module path in docstring**

The docstring example still references the old module path `score_metamodel.sphinx_filters.generic_pie_linked_items(...)` — this module was moved to `score_metrics`. It should be `score_metrics.sphinx_filters.generic_pie_linked_items(...)`. Any user copying from this example will get an import error at runtime.

AlexanderLanin (commented):

**Critical: 0, Important: 2, Suggestions: 1**

---

This pass covers `score_metrics` (the new extension), the CI workflow rewrite, and cross-cutting consistency. The autodiscovery design in `score_metamodel` is already well-covered by the existing review comments.

**Core issue — `score_metamodel_yaml` coupling is unguarded.** `score_metrics/traceability_metrics.py:131` does `app.config.score_metamodel_yaml`, but that config value is registered by `score_metamodel.setup()`, not `score_metrics`. In the bundle (`score_sphinx_bundle`) `score_metamodel` is loaded first, so it happens to work — but `score_metrics` is documented as a standalone extension and its `setup()` does not declare this dependency. A consumer loading only `score_metrics` will hit `AttributeError`. At minimum, `score_metrics` should either register `score_metamodel_yaml` itself (with the same default `""`) or document the hard dependency on `score_metamodel`.

**Tag mismatch — inline comment on `traceability_metrics.py:141`.** The config description says autodiscovery uses `{"requirement", "requirement_excl_process"}`, but the code passes only `{"requirement"}`. Types tagged exclusively with `requirement_excl_process` are silently excluded. Doc and code need to agree on which tags drive discovery.

**AI-generated docs — inline comment on `dashboards_and_quality_gates.rst:156`.** ~295 lines of Copilot-generated API reference content were committed into a user-facing how-to guide. It documents internal functions and implementation details. Should be removed.

**Minor: stale module path in docstring** — inline comment on `sphinx_filters.py`. The `:filter-func:` example still says `score_metamodel.sphinx_filters.*`, which no longer exists.

**No concern** with the CI workflow rewrite — the simplification is well-motivated and the parametrized test matrix is cleaner. The removal of `--lockfile_mode=error` from `bazel run //:ide_support` is worth a quick confirmation that it was intentional.

src/extensions/score_metrics/sphinx_filters.py
MaximilianSoerenPollak: Fixed

src/extensions/score_metrics/__init__.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    if exception:
+        logger.error(f"Sphinx-Exception at end of build: {exception}")
+    out_path = Path(app.outdir) / "metrics.json"
+    out_path.parent.mkdir(parents=True, exist_ok=True)
+    out_path.write_text(json.dumps(CALCULATED_METRICS, indent=2), encoding="utf-8")
+    print(f"Traceability metrics written to: {out_path}")
+
+
+def setup(app: Sphinx) -> dict[str, str | bool]:
```
AlexanderLanin, line 53: _AI-assisted review (Claude):_

**Important — `score_metamodel_requirement_types` is registered by `score_metrics`, not `score_metamodel`**

`score_metrics/__init__.py` registers `score_metamodel_requirement_types` and `score_metamodel_include_external_needs` as config values, then calls `app.setup_extension("score_metamodel")`. This means:

1. The config values are named `score_metamodel_*` but owned by `score_metrics` — confusing and semantically wrong.
2. If `score_metamodel` is loaded before `score_metrics` (e.g. standalone, without the bundle), `score_metamodel_yaml` is registered but these two keys are not, making the pairing asymmetric.
3. Any consumer who loads both extensions independently could hit a duplicate `add_config_value` registration error.

These config keys should either be renamed to `score_metrics_*` to reflect actual ownership, or `score_metamodel` should register them and `score_metrics` should just read them.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -40,20 +40,34 @@ Typical Setup
...
 
-Minimal Configuration Example
------------------------------
+Configuration
+-------------
+
+Default Behavior (No Configuration Needed)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, ``score_metamodel`` autodiscovers requirement types from the
```
AlexanderLanin: _AI-assisted review (Claude):_

**Important — docs still say `score_metamodel` writes `metrics.json`**

Lines 49, 77 (and surrounding context) say:
> "By default, `score_metamodel` autodiscovers requirement types..."
> "The documentation build writes `metrics.json` via `score_metamodel`..."

But after this PR, it is `score_metrics` that writes `metrics.json`. `score_metamodel` does not have this role anymore. Any consumer following these docs will be confused about which extension controls the metric output.

scripts_bazel/traceability_gate.py
```diff
@@ -119,10 +148,53 @@ def _check_type_thresholds(
...
+
+    Returns:
+        Process exit code:
+        - 0 on success
+        - 1 on input or schema errors
+        - 2 on threshold failures
+    """
     parser = argparse.ArgumentParser(
         description=(
             "Read a traceability metrics JSON (schema v1) and enforce coverage "
```
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion — argparse description says "schema v1" but gate enforces schema v2**

```python
description=(
    "Read a traceability metrics JSON (schema v1) and enforce coverage "
    ...
```

`_SUPPORTED_SCHEMA_VERSION = "2"` — this will appear verbatim in `--help` output and confuse users running the gate.

```suggestion
description=(
    "Read a traceability metrics JSON (schema v2) and enforce coverage "
    ...
```

AlexanderLanin (commented):

**Re-review pass — Critical: 0, Important: 3, Suggestions: 1**

---

## Comment resolution status

✅ Fixed since last review:
- `_req` suffix fallback — removed
- `_get_need_value` / `except Exception` over-defensiveness — removed
- `docs/conf.py` explicit config noise — removed
- `_as_requirement_directive` pointless abstraction — removed
- Stale `score_metamodel.sphinx_filters` module path in docstring — fixed

❌ Still open:
- **Tag mismatch** (`{"requirement"}` vs documented `{"requirement", "requirement_excl_process"}`) — not fixed
- **AI-generated Copilot block** in `dashboards_and_quality_gates.rst` — still present at line ~155
- **`score_metamodel_yaml` implicit dependency** — `score_metrics` reads `app.config.score_metamodel_yaml` but does not register it; relies on `app.setup_extension("score_metamodel")` being called first

---

## Architecture summary

The split is the right direction: `score_metamodel` → metamodel/validation, `score_metrics` → metric collection + CI gate output. The deletion of the old `traceability_metrics.py` from `score_metamodel` is clean. No live duplication.

Three structural issues with the current state:

**1. Config key ownership mismatch (inline comment on `score_metrics/__init__.py:53`).**
`score_metamodel_requirement_types` and `score_metamodel_include_external_needs` are registered by `score_metrics`, not `score_metamodel`. Names imply one owner, code says another. Risk of duplicate registration if both extensions are loaded independently.

**2. Docs attribute metrics.json to the wrong extension (inline comment on `dashboards_and_quality_gates.rst:49`).**
Multiple lines say `score_metamodel` writes `metrics.json`. It does not — `score_metrics` does. Anyone configuring or troubleshooting metrics generation will look in the wrong place.

**3. Schema v2 is a silent breaking change for existing consumers.**
The gate now rejects schema v1 with a hard error. There is no mention of this in the docs, no migration guide, and no CHANGELOG entry. Any downstream consumer running the old gate against a new build (or vice versa) will get `unsupported schema_version 1` with no upgrade path. This should at minimum be called out in the how-to guide.

**4. Argparse description says "schema v1" (inline comment on `traceability_gate.py:200`).**
One-line fix, shows up in `--help`.

---

## Gate correctness

The gate logic itself is sound for the common case. One edge case worth noting: `safe_percent(0, 0)` returns `100.0`, so a build with zero testcases will report `linked_to_requirements_pct = 100.0` and pass any `--min-tests-linked` threshold. The gate does guard the empty `metrics_by_type` case correctly, but not the zero-tests case. Whether this is acceptable behavior is a product decision — if intentional, it should be documented.

MaximilianSoerenPollak:
Answer to clanker review: 

I adapted the documentation to reflect the right extension now. 

Score_metamodel will always be loaded as well so it 'theoretically' should always work now that I have entered it as a requirement in score_metrics. 

I do not see a reason to mention that v2 breaks schema v1. Nobody used it yet / we didn't roll it out or helped implement it in repos. 

The gate correctness (0/0 = 100%) is something I'm unsure if this is good, or just doesn't make sense anymore. Not quiet sure.

AlexanderLanin (commented):

**Re-review — 2026-06-08**

Most prior comments have been addressed. Two blockers remain before merge.

---

**Critical — config key name mismatch: code reads the wrong key**

`score_metrics/__init__.py` registers:
- `score_metrics_requirement_types` (line 55)
- `score_metrics_include_external_needs` (line 65)

But the code that actually reads them uses the **old** names:

- `traceability_metrics.py:138` reads `score_metamodel_requirement_types`
- `score_metrics/__init__.py:36` reads `score_metamodel_include_external_needs`

These are `getattr(..., default)` calls, so they **silently return the default** when a consumer sets the new `score_metrics_*` key. The explicit override never takes effect. Autodiscovery will still work (empty string → autodiscover), but any consumer that set the config to customize requirement types or enable external needs will get wrong metrics with no error.

Fix: change both `getattr` reads to use the `score_metrics_*` names that are actually registered.

---

**Important — Copilot-generated content still committed in user-facing docs**

`docs/how-to/dashboards_and_quality_gates.rst` still contains the `╓ Some portions generated by Github Copilot ╜` banner (line ~155). The section below it ("Needpie Usage Guide") now looks cleaned up, but the banner is still there and should be removed — it has no place in shipped documentation.

---

Critical: 1, Important: 1

AlexanderLanin:
rst is not "shipped" so comment can stay

AlexanderLanin (commented):

**Re-review — 2026-06-08 (updated)**

The two blockers from my last comment are resolved:

- ✅ Config key mismatch fixed — `traceability_metrics.py` and `score_metrics/__init__.py` now consistently read `score_metrics_requirement_types` / `score_metrics_include_external_needs`.
- ✅ Copilot banner — acknowledged as not shipped, accepted as-is.
- ✅ `requirement_excl_process` tag concern is a non-issue: every type tagged `requirement_excl_process` in the metamodel also carries `requirement`, so autodiscovery with `{"requirement"}` covers all of them.
- ✅ Docs updated to reference `score_metrics` throughout and use the correct config keys.

No remaining blockers. LGTM from my side.

AlexanderLanin (approved):

AlexanderLanin:
I think this has been discussed at length and it's time to merge. Sure, further changes will be required. But those are hopefully small and can happen on main.

MaximilianSoerenPollak:
@AlexanderLanin  I think what is next, is now to build around this and integrate the CI gates?

PR #549, opened 2026-05-21
docs: rewrite test_to_doc_links how-to for C++, Rust, and Python

a-zw:
See https://eclipse-score.github.io/docs-as-code/pr-549/how-to/test_to_doc_links.html

### DIFF
--- a/docs/how-to/test_to_doc_links.rst
+++ b/docs/how-to/test_to_doc_links.rst
@@ -17,47 +17,76 @@
 Reference Docs in Tests
 =======================
 
-In tests, you want to reference requirements (needs).
-The docs-as-code tool will create backlinks in the documentation.
+This guide explains how to annotate test cases so that
+docs-as-code automatically creates traceability links between tests and requirements.
 
-Docs-as-code parses `test.xml` files produced by Bazel under `bazel-testlogs/`.
-To attach metadata to tests use the project tooling decorator (provided by the
-attribute plugin). Example usage:
+The mechanism is language-agnostic:
+Bazel produces JUnit-style ``test.xml`` files under ``bazel-testlogs/``.
+The ``score_source_code_linker`` extension parses those files, extracts test metadata
+(name, file, line, result, and verification properties),
+and creates backlinks on the referenced requirements.
 
-.. code-block:: python
+Required Properties
+-------------------
 
-	 from attribute_plugin import add_test_properties
+Every linked test must declare the following properties
+(see :need:`gd_guidl__verification_specification`):
 
-	 @add_test_properties(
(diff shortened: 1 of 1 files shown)


docs/how-to/test_to_doc_links.rst
```diff
@@ -17,47 +17,76 @@
...
-	 from attribute_plugin import add_test_properties
+Every linked test must declare the following properties
+(see :need:`gd_guidl__verification_specification`):
 
-	 @add_test_properties(
-			 partially_verifies=["tool_req__docs_common_attr_title", "tool_req__docs_common_attr_description"],
-			 test_type="interface-test",
-			 derivation_technique="boundary-values",
-	 )
-	 def test_feature():
```
AlexanderLanin, line 36: _AI-assisted review (Claude):_

**Important** — `Description` is listed as a mandatory property for link creation, but the linker does not enforce it.

In `xml_parser.py:160`, the `Description` property is explicitly skipped during parsing:
```python
if prop_name == "Description":
    continue
```

And `DataOfTestCase` in `testlink.py` has no `Description` field — so `is_valid()` never checks for it. A test that omits `Description` will still be linked successfully.

The statement on line 47 — *"If any mandatory property is missing the test will be skipped during link creation"* — is technically incorrect for `Description`. It applies to `PartiallyVerifies`/`FullyVerifies`, `TestType`, and `DerivationTechnique`, but not to `Description`.

`Description` may well be a *process* requirement (from the verification specification), but the way it's presented here implies the linker enforces it, which it doesn't.

**Suggested fix:** Either clarify that `Description` is a process requirement (not enforced by the linker), or split the list into linker-enforced vs. process-required properties.

AlexanderLanin (commented):

**Severity breakdown:** Critical: 0, Important: 1, Suggestions: 1

Nice rewrite — the structure is much clearer and the language-specific guidance with cross-references to requirements is a solid improvement over the old Python-only example.

**One accuracy issue with the `Description` property claim** (see inline comment) and **one minor discoverability note** about a dropped cross-reference to `dashboards_and_quality_gates`. The old doc had a "Related" section linking there — might be worth keeping as a brief "See also" since it covers end-to-end dashboard and CI threshold setup, which is a natural next step after setting up test links.

The two external need references (`gd_guidl__verification_specification`, `gd_temp__verification_specification`) appear to resolve correctly in the PR preview build, so those look fine.

docs/how-to/test_to_doc_links.rst
a-zw: I consider that a bug in our implementation but this documentation is fine.

AlexanderLanin (approved):

PR #552, opened 2026-05-22
feat: support no-java repos

AlexanderLanin:
Currently docs cannot be used without Java. For repos that do not use any uml, this is just overhead that is downloaded all the time, but never actually used.

Downside: potential issues when such repos do want to use plantuml and wonder why its not working.


AlexanderLanin:
Moving to draft as this seems to be a rare use case

PR #555, opened 2026-05-22
feat: force clean build after warnings or dependency changes

AlexanderLanin:
## Why

When `bazel run //:docs` (or any sibling target) runs Sphinx and the build fails, the `_build/` directory is left on disk. On the next invocation Sphinx rebuilds incrementally and skips unchanged files — so the warnings that caused the failure simply **disappear from the output** without the underlying problem being fixed. This makes it easy to silently miss warnings across multiple invocations.

A second, related problem: when `MODULE.bazel`, `MODULE.bazel.lock`, or the root `BUILD` file changes (new dependency, version bump, build graph change), `_build/` is stale but Sphinx has no way of knowing — it will keep serving output built against the old dependency tree.

## What

**Dirty-build detection via `warnings.txt`**

Sphinx's native `--warning-file` flag is used to capture all warnings into `_build/warnings.txt`. At the start of each build, if that file exists and is non-empty, `_build/` is deleted before Sphinx runs. This guarantees that every warning is always visible and that a clean build is performed until the issue is resolved.

**Sentinel-file hash check**

A SHA-256 hash over the three files most likely to invalidate the build graph (`MODULE.bazel`, `MODULE.bazel.lock`, `BUILD`) is stored in `_build/.module_bazel_hash` after every successful build. On the next invocation the stored hash is compared to the current one; a mismatch causes `_build/` to be deleted.

Both checks are handled by a single function `clean_builddir_if_stale` so the logic stays in one place.

Only the non-interactive actions (`incremental`, `check`, `linkcheck`) are affected. `live_preview` (sphinx-autobuild) is unchanged.

## Changes

- `src/incremental.py` — `clean_builddir_if_stale`, `update_module_hash`, and `_compute_hash` added as module-level functions; `--warning-file` added to Sphinx arguments; `--keep-going` preserved; hash written on successful build, separator appended to `warnings.txt` on failure for easier debugging
- `src/incremental_dirty_build_test.py` — new test file, 7 focused unit tests using pyfakefs (no subprocess, no real filesystem)
- `src/BUILD` — test target added
- `src/requirements.in` / `requirements.txt` — `pyfakefs` added

### DIFF
--- a/src/incremental.py
+++ b/src/incremental.py
@@ -48,17 +48,22 @@ def _compute_hash(files: list[Path]) -> str:
 
 def clean_builddir_if_stale(build_dir: Path, sentinel_files: list[Path]) -> None:
     """Delete build_dir if the previous build had warnings or any sentinel file changed."""
+    if not build_dir.exists():
+        return
+
     warnings_txt = build_dir / "warnings.txt"
     has_warnings = warnings_txt.exists() and warnings_txt.stat().st_size > 0
 
     hash_file = build_dir / _MODULE_HASH_FILE
     hash_changed = (
-        hash_file.exists()
-        and hash_file.read_text().strip() != _compute_hash(sentinel_files)
+        not hash_file.exists()
+        or hash_file.read_text().strip() != _compute_hash(sentinel_files)
     )
 
     if has_warnings or hash_changed:
-        print("Previous build had warnings or the hash changed. Removing _build to ensure a clean build.")
+        print(
+            "Previous build had warnings or the hash changed. Removing _build to ensure a clean build."
+        )
         shutil.rmtree(build_dir)
 
 
--- a/src/incremental.py
+++ b/src/incremental.py
@@ -12,8 +12,10 @@
 # *******************************************************************************
 
 import argparse
+import hashlib
 import logging
 import os
+import shutil
 import sys
 import time
 from pathlib import Path
@@ -26,6 +28,8 @@ from sphinx_autobuild.__main__ import (
 
 logger = logging.getLogger(__name__)
 
+_MODULE_HASH_FILE = ".module_bazel_hash"
+
 
 def get_env(name: str) -> str:
     val = os.environ.get(name, None)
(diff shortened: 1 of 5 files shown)


src/incremental.py
```diff
@@ -35,6 +39,34 @@ def get_env(name: str) -> str:
...
+    hash_file = build_dir / _MODULE_HASH_FILE
+    hash_changed = (
+        hash_file.exists()
+        and hash_file.read_text().strip() != _compute_hash(sentinel_files)
+    )
+
+    if has_warnings:
+        print("Previous build had warnings. Removing _build to ensure a clean build.")
+    if has_warnings or hash_changed:
+        shutil.rmtree(build_dir)
```
MaximilianSoerenPollak: Nit: 

A bit weird the logic here, imo. 

Wouldn't be this not better?

```suggestion
    if has_warnings or hash_changed:
        print("Previous build had warnings or the hash changed. Removing _build to ensure a clean build.")
        shutil.rmtree(build_dir)
```

src/incremental.py
```diff
@@ -73,9 +105,21 @@ def get_env(name: str) -> str:
     else:
         workspace = ""
 
+    build_dir = Path(workspace + "_build")
+    sentinel_files = [
```
MaximilianSoerenPollak, line 113: Just curious bout the naming, is it meaing like `we watch these files` therefore `sentinel` ?

src/incremental.py
```diff
@@ -35,6 +39,34 @@ def get_env(name: str) -> str:
...
+def clean_builddir_if_stale(build_dir: Path, sentinel_files: list[Path]) -> None:
+    """Delete build_dir if the previous build had warnings or any sentinel file changed."""
+    warnings_txt = build_dir / "warnings.txt"
+    has_warnings = warnings_txt.exists() and warnings_txt.stat().st_size > 0
+
+    hash_file = build_dir / _MODULE_HASH_FILE
+    hash_changed = (
+        hash_file.exists()
+        and hash_file.read_text().strip() != _compute_hash(sentinel_files)
+    )
```
MaximilianSoerenPollak, line 61: What happens if hash_file does not exists?

src/incremental.py
MaximilianSoerenPollak: Also if I remember correctly the warnings file will always exist right?

MaximilianSoerenPollak (commented):
Some questions

src/incremental.py
AlexanderLanin: Thats why `has_warnings` checks for file size > 0.

src/incremental.py
AlexanderLanin: yeah exactly. Not watch, but these files will trigger a change. So they watch/protect/whatever the build.

src/incremental.py
AlexanderLanin: Currently not exist = no rebuild. We could actually change that to be very sure...

MaximilianSoerenPollak: That is fine by me I think.

AlexanderLanin: _AI fix-commit (Claude):_ `f727650d` — treat missing hash file as stale; add early return for absent build_dir; add `test_missing_hash_file_triggers_clean`

AlexanderLanin: fixed

MaximilianSoerenPollak (approved):
Seems good for me for now. 
We might can expand tests in the future, but that is something we can look at once we have time...

PR #556, opened 2026-05-26
fix: use label for robustness

a-zw:
## 📌 Description

``Label()`` is the best way to specify targets "relative to lexically-here". For contrast, something like "//foo:bar" is "relative to dynamically-here" and fails if one executes rules from another module. It is mostly equivalent to "@my_module//foo:bar" but avoids repeating the current module name. It also works when module names are remapped (but afaik nobody does that yet in S-CORE).

Also, fix two occasions where code might break if labels are involved.

See https://bazel.build/rules/lib/builtins/Label for context.

## 🚨 Impact Analysis

## ✅ Checklist


AlexanderLanin:
@a-zw can you explain why this is better?

MaximilianSoerenPollak:
> @a-zw can you explain why this is better?

He specified this in another chat: 

>Label() is the best way to specify targets "relative to lexically-here". For contrast, something like "//foo:bar" is "relative to dynamically-here" and fails if one executes rules from another module. It is mostly equivalent to "@my_module//foo:bar" but avoids repeating the current module name. It also works when module names are remapped (but afaik nobody does that yet in S-CORE).

AlexanderLanin:
So the argument is `Label` is more readable then `@score_docs_as_code`?

a-zw:
@MaximilianSoerenPollak I added that to the description (assuming that it ends up in the merge commit).

@AlexanderLanin It is not about readability but about robustness. If we rename the module, Label just keeps working. If someone renames the module when importing it, Label just keeps working.

It is not a critical thing because I don't see any of these things happening in the near future. It looks like a good practice though.

AlexanderLanin:
> assuming that it ends up in the merge commit)

It links to this PR anyway, so thats fine.

AlexanderLanin (approved):

PR #560, opened 2026-05-27
feat: Provide :metrics_json target

a-zw:
With this Bazel rules can depend on exactly that file instead of somehow extracting it from :needs_json in hacky ways.

This seems to be the cleanest way, because we cannot change the sphinx_docs rule to add additional special outputs.


AlexanderLanin (approved):

MaximilianSoerenPollak (approved):
At some point we probably should split up docs.bzl it's getting so messy and big.

PR #561, opened 2026-05-27
test: avoid chdir by passing git_root

AlexanderLanin:
`chdir` is a major driver for inter-test dependencies. One order of tests works, another does not. This PR avoid some easy cases of chdir by passing `git_root` instead of dynmically finding it based on cwd.

### DIFF
--- a/src/extensions/score_source_code_linker/helpers.py
+++ b/src/extensions/score_source_code_linker/helpers.py
@@ -35,7 +35,8 @@ def get_github_link(
     """
     Get GitHub link for a file and line number using git information from the local repository.
 
-    @param git_root: Optional path to the git repository root. If not provided, it will be auto-detected.
+    Args:
+        git_root: Optional path to the git repository root. If not provided, it will be auto-detected.
     """
     if link is None:
         link = DefaultNeedLink()
--- a/src/extensions/score_source_code_linker/helpers.py
+++ b/src/extensions/score_source_code_linker/helpers.py
@@ -30,26 +30,32 @@ from src.helper_lib import (
 def get_github_link(
     metadata: RepoInfo,
     link: NeedLink | DataForTestLink | DataOfTestCase | None = None,
+    git_root: Path | None = None,
 ) -> str:
+    """
+    Get GitHub link for a file and line number using git information from the local repository.
+
+    @param git_root: Optional path to the git repository root. If not provided, it will be auto-detected.
+    """
     if link is None:
         link = DefaultNeedLink()
     if not metadata.hash:
+        if not git_root:
+            git_root = find_git_root() or Path()
         # Local path (//:docs)
-        return get_github_link_from_git(link)
+        return get_github_link_from_git(git_root, link)
     # Ref-Integration path (//:docs_combo..)
     return get_github_link_from_json(metadata, link)
 
 
 def get_github_link_from_git(
+    git_root: Path,
     link: NeedLink | DataForTestLink | DataOfTestCase | None = None,
 ) -> str:
     if link is None:
         link = DefaultNeedLink()
-    passed_git_root = find_git_root()
-    if passed_git_root is None:
-        passed_git_root = Path()
-    base_url = get_github_base_url()
-    current_hash = get_current_git_hash(passed_git_root)
+    base_url = get_github_base_url(git_root)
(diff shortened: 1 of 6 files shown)


AlexanderLanin (commented):

**Critical: 0 | Important: 0 | Suggestions: 2**

Clean, well-scoped change. The core refactoring is correct: `git_root` is now passed explicitly instead of being resolved from `cwd`, which eliminates the inter-test `chdir` dependency for these specific call paths. The `git_root: Path | None = None` default on `get_github_link` keeps the production caller in `__init__.py` (which does not pass `git_root`) working transparently.

Two minor issues worth fixing:

**1. Stale docstring in `get_github_base_url` (`src/helper_lib/__init__.py`, line 141)**

The docstring still reads:
> `bazel build => ⚠️ Uses Path() fallback when git_root is None`

But `git_root` is now a required `Path` — there is no `None` case or `Path()` fallback in this function anymore. That line is now misleading.

**2. Extra blank line in `test_source_code_link_integration.py` (line ~215)**

Removing `construct_gh_url` left a double blank line (`\n\n\n`) before the `@pytest.fixture()`. PEP 8 / flake8 flags this. The `# flake8: noqa: E501` suppression at line 213 only covers line length, not E303.

**Broader context:** Several other `os.chdir` calls remain in the test suite (e.g. `test_consumer.py:556`, `test_helper_lib.py:167,190`, `test_source_code_link_integration.py:233`). The PR description says this is intentional scoping ("some easy cases"), which is fine — incremental improvement is a valid approach here.

src/extensions/score_source_code_linker/helpers.py
```diff
@@ -30,26 +30,32 @@
 def get_github_link(
     metadata: RepoInfo,
     link: NeedLink | DataForTestLink | DataOfTestCase | None = None,
+    git_root: Path | None = None,
 ) -> str:
+    """
+    Get GitHub link for a file and line number using git information from the local repository.
+
+    @param git_root: Optional path to the git repository root. If not provided, it will be auto-detected.
```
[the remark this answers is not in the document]
AlexanderLanin: _AI fix-commit (Claude):_ `e67ac539` — switch @param to Args: style to match project convention

AlexanderLanin:
_AI fix-commit (Claude):_ `f635f762` — stale docstring + extra blank line fixed; `e67ac539` — @param → Args: style

src/extensions/score_source_code_linker/tests/test_source_code_link_integration.py
```diff
@@ -213,11 +213,6 @@ def make_test_xml_2():
 # flake8: noqa: E501 (end)
 
 
-def construct_gh_url() -> str:
-    gh = get_github_base_url()
-    return f"{gh}/blob/"
```
MaximilianSoerenPollak, line 218: Why is this deleted without compensation? 
Was this not used?

src/helper_lib/__init__.py
```diff
@@ -132,22 +132,19 @@ def get_github_repo_info(git_root_cwd: Path) -> str:
...
 
-def get_github_base_url() -> str:
+def get_github_base_url(git_root: Path) -> str:
     """
     Generate GitHub base URL for the current repository.
 
     Execution context behavior:
     - 'bazel run' => ✅ Correct GitHub URL
-    - 'bazel build' => ⚠️ Uses Path() fallback when git_root is None
+    - 'bazel build' => ⚠️ git_root must be provided by the caller; no auto-detection here
```
MaximilianSoerenPollak, line 141: So `git_root` could theoretically also be just `Path()` right?

src/extensions/score_source_code_linker/tests/test_source_code_link_integration.py
AlexanderLanin: nope, it wasnt 😆

src/helper_lib/__init__.py
AlexanderLanin: Yes. I did not change the logic in this PR. The fallback to Path() was already implemented.

MaximilianSoerenPollak (approved):
Looks good. 

This should improve consitency of bazel & pytest behaviour.

PR #562, opened 2026-05-28
test: avoid changing os.environ

AlexanderLanin:
## Why

Tests that directly assign to `os.environ` and clean up manually are fragile: if a test raises before the cleanup code runs, the environment variable leaks into every subsequent test in the same process. Because several of these fixtures were scoped to `session`, a leaked `BUILD_WORKSPACE_DIRECTORY` or `KNOWN_GOOD_JSON` could silently influence unrelated tests, making failures order-dependent and hard to reproduce. The manual try/finally pattern for saving and restoring the old value is also noisy boilerplate that obscures the test's actual intent.

A subtler problem with `session`-scoped fixtures is that `monkeypatch` only supports `function` and `module` scope — you can't use it with `session`. The fix must therefore reduce the fixture scope to `module` before switching to `monkeypatch`.

## What

Replace all direct `os.environ` mutations in tests with `monkeypatch.setenv`. This is pytest's standard mechanism for environment isolation: it automatically reverts every change after the test (or module) finishes, even if the test raises. No manual save/restore or `os.environ.pop` calls are needed.

Fixtures in `test_repo_source_link_integration.py` and `test_source_code_link_integration.py` that were previously `session`-scoped are downgraded to `module`-scoped so that `monkeypatch` can be used safely. The `git_repo_setup` fixtures no longer set `BUILD_WORKSPACE_DIRECTORY` themselves; each test that needs it sets it via `monkeypatch.setenv` instead, keeping environment side-effects visible at the call site.

In `test_helper_lib.py`, the `os.chdir` call in `test_runfiles_dir_found` is also removed — it was unnecessary because `get_runfiles_dir()` reads the env var directly and never consults the working directory.

## Changes

- `test_repo_source_link_integration.py` — downgrade `sphinx_base_dir`, `git_repo_setup`, `create_demo_files` from `session` to `module` scope; remove `os.environ` assignment from `git_repo_setup`; replace six direct `os.environ["BUILD_WORKSPACE_DIRECTORY"] = …` assignments with `monkeypatch.setenv`
- `test_source_code_link_integration.py` — same scope downgrade; remove `os.environ` assignment from `git_repo_setup`; replace two direct assignments with `monkeypatch.setenv`
- `test_xml_parser.py` — replace two manual try/finally save-restore patterns around `KNOWN_GOOD_JSON` with `monkeypatch.setenv`
- `test_helper_lib.py` — replace manual `os.environ` mutation and `os.environ.pop` cleanup with `monkeypatch.setenv`; remove the unnecessary `os.chdir` call

### DIFF
--- a/src/extensions/score_source_code_linker/tests/test_repo_source_link_integration.py
+++ b/src/extensions/score_source_code_linker/tests/test_repo_source_link_integration.py
@@ -45,12 +45,12 @@ from src.extensions.score_source_code_linker.repo_source_links import (
 """
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def sphinx_base_dir(tmp_path_factory: TempPathFactory) -> Path:
     return tmp_path_factory.mktemp("test_module_links_repo")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
 def git_repo_setup(sphinx_base_dir: Path) -> Path:
     """Creating git repo, to make testing possible"""
     repo_path = sphinx_base_dir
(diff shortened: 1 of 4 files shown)


MaximilianSoerenPollak (approved):
Ye that looks good to me. 
Makes sense ot use the monkeypatch for the thing it actually was meant to be used.

PR #563, opened 2026-05-28
test: remove more chdir calls

AlexanderLanin:
## Why

`os.chdir` in tests changes the working directory of the entire process. When a test raises or is interrupted before restoring the original directory, every subsequent test in the session inherits the wrong CWD. This causes false failures that are order-dependent and hard to reproduce — especially in Bazel sandboxes where the original directory may no longer exist at all.

The `sphinx_app_setup` fixtures in the two source-code-linker integration test modules worked around this with a `contextlib.suppress(FileNotFoundError)` guard and a try/finally restore — the suppress and the stale comment "Original directory might not exist anymore in Bazel sandbox" are evidence of exactly this failure mode being hit before. `test_helper_lib.py` had two tests calling `os.chdir` for no reason (the functions under test don't consult CWD), and `test_rules_file_based.py` used `os.chdir` instead of the pytest-provided equivalent.

## What

Replace all remaining `os.chdir` calls with `monkeypatch.chdir`. pytest's `monkeypatch` reverts the directory after each test, even on failure, making the cleanup unconditional and removing the need for try/finally guards.

In `sphinx_app_setup` the try/finally + `contextlib.suppress` pattern is removed entirely: `monkeypatch.chdir` is called once before `SphinxTestApp` construction and the fixture becomes a straight-line function. Two unused imports (`contextlib`, `os`) are removed as a result.

In `test_helper_lib.py`, the two `os.chdir` calls are dropped outright — `get_current_git_hash` and `get_repo_info` both receive an explicit `git_dir` argument and never read CWD, so the chdir was dead code.

## Changes

- `test_repo_source_link_integration.py` — replace try/finally `os.chdir` in `sphinx_app_setup` with `monkeypatch.chdir`; remove `contextlib` and `os` imports
- `test_source_code_link_integration.py` — same: replace the Bazel-workaround try/finally pattern with `monkeypatch.chdir`; remove stale comments, `contextlib`, and `os` imports
- `test_helper_lib.py` — drop two `os.chdir` calls from `test_git_operations_with_no_commits` and `test_git_repo_with_no_remotes`, which were unnecessary because the functions under test take an explicit path argument
- `test_rules_file_based.py` — replace `os.chdir(app.srcdir)` with `monkeypatch.chdir(app.srcdir)`; remove `os` import


src/helper_lib/test_helper_lib.py
```diff
@@ -164,7 +166,6 @@ def test_git_operations_with_no_commits(temp_dir: Path):
     )
     subprocess.run(["git", "config", "user.name", "Test User"], cwd=git_dir, check=True)
 
-    os.chdir(Path(git_dir).absolute())
```
MaximilianSoerenPollak, line 167: This gets no monkeypatch? 
Is that correct?

AlexanderLanin: Haven't you read the PR description 😲

AlexanderLanin: > In test_helper_lib.py, the two os.chdir calls are dropped outright — get_current_git_hash and get_repo_info both receive an explicit git_dir argument and never read CWD, so the chdir was dead code.

MaximilianSoerenPollak (approved):

PR #564, opened 2026-05-28
Fix: Update RST File Based Tests

MaximilianSoerenPollak:
## 📌 Description
Current RST tests were quite hard to code directly and were error prone without having any insight. 
This PR changes the RST tests to make them more explicit as well as making future expandability easier possible

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs/internals/extensions/rst_filebased_testing.md
+++ b/docs/internals/extensions/rst_filebased_testing.md
@@ -19,8 +19,8 @@ The test files are expected to contain the following format:
 
     #CHECK: <check functions>
 
-    #EXPECT: <warning message>
-    #EXPECT-NOT: <warning message>
+    #EXPECT[+x]: <warning message>
+    #EXPECT-NOT[+x]: <warning message>
 
     <need information>
 
@@ -36,6 +36,9 @@ Message text which is expected/not expected during the
                     This message is checked for the Sphinx-Needs directive
                     specified after the EXPECT/EXPECT-NOT statement.
 
+This message needs a '[+x]'offset after the 'EXPECT/-NOT' that should point to the need 
+that should (not) emit the warning.
+
 **\<need information>**<br>
 One or more Sphinx-Needs directives needed for the
                     Sphinx document build
@@ -43,7 +46,7 @@ One or more Sphinx-Needs directives needed for the
 **Example:**
 
     #CHECK: check_options
-    #EXPECT: std_wp__test__abcd: is missing required attribute: `status`.
+    #EXPECT[+2]: std_wp__test__abcd: is missing required attribute: `status`.
 
     .. std_wp:: Test requirement
         :id: std_wp__test__abcd
@@ -51,3 +54,19 @@ One or more Sphinx-Needs directives needed for the
 This example verifies that the warning message
 *std_wp__test__abcd: is missing required attribute: \`status\`*
 is shown during the Sphinx build. Only the *check_options* check is enabled.
+
+With the '[+2]' after the 'EXPECT' we tell the parser that we want this warning 
+to be emitted and checked for 2 lines underneath
+
+There is multiple things that are not allowed for example, you need to have 
+a new line between the EXPECT/-NOT and the need that it refers to
+
+**Negative Example**
+
+
+    #EXPECT-NOT[+1]: std_wp__test__abcd: is missing required attribute: `status`.
+    .. std_wp:: Test requirement
+        :id: std_wp__test__abcd
+
+This will error and let you know that an offset of '1' is not allowed and you 
+need to add a new line beneath the Warning Statement
--- a/docs/internals/extensions/rst_filebased_testing.md
+++ b/docs/internals/extensions/rst_filebased_testing.md
@@ -36,7 +36,7 @@ Message text which is expected/not expected during the
                     This message is checked for the Sphinx-Needs directive
                     specified after the EXPECT/EXPECT-NOT statement.
 
-This message needs a '[+x]'offset after the 'EXPECT/-NOT' that should point to the need 
+This message needs a '[+x]'offset after the 'EXPECT/-NOT' that should point to the need
 that should (not) emit the warning.
 
 **\<need information>**<br>
@@ -55,10 +55,10 @@ This example verifies that the warning message
 *std_wp__test__abcd: is missing required attribute: \`status\`*
 is shown during the Sphinx build. Only the *check_options* check is enabled.
 
-With the '[+2]' after the 'EXPECT' we tell the parser that we want this warning 
+With the '[+2]' after the 'EXPECT' we tell the parser that we want this warning
 to be emitted and checked for 2 lines underneath
 
-There is multiple things that are not allowed for example, you need to have 
+There is multiple things that are not allowed for example, you need to have
 a new line between the EXPECT/-NOT and the need that it refers to
 
 **Negative Example**
@@ -68,5 +68,5 @@ a new line between the EXPECT/-NOT and the need that it refers to
     .. std_wp:: Test requirement
         :id: std_wp__test__abcd
 
-This will error and let you know that an offset of '1' is not allowed and you 
+This will error and let you know that an offset of '1' is not allowed and you
 need to add a new line beneath the Warning Statement
--- a/src/extensions/score_metamodel/tests/test_rules_file_based.py
+++ b/src/extensions/score_metamodel/tests/test_rules_file_based.py
@@ -143,11 +143,10 @@ def extract_test_data(rst_file: Path) -> tuple[RstData, list[ErrorChecks]]:
     rst_data = RstData(filename=str(rst_file.relative_to(RST_DIR)))
     parsed_checks: list[ErrorChecks] = []
     with open(rst_file) as f:
-        for no, raw_line in enumerate(f, start=1):
-            line = raw_line.strip() 
+        for no, line in enumerate(f, start=1):
             # Beginning of new need
             # We filter for '::' as well so we ONLY get directives not comments
-            if line.startswith(".. ") and "::" not in line:
+            if line.startswith(".. ") and "::" in line:
                 rst_data.found_objects.append(no)
                 continue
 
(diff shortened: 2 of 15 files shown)


src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -20,28 +19,24 @@
 
 import pytest
 from sphinx.testing.util import SphinxTestApp
+from sphinx_needs.logging import get_logger
 
 RST_DIR = Path(__file__).absolute().parent / "rst"
 DOCS_DIR = Path(__file__).absolute().parent.parent.parent.parent.parent
```
AlexanderLanin: _AI-assisted review (Claude):_

`DOCS_DIR` is defined but no longer used anywhere in the file after the `shutil.copytree` was removed from `sphinx_base_dir`. Same for `logger` — it is imported and instantiated at module level but never called. Both should be deleted.

```suggestion
RST_DIR = Path(__file__).absolute().parent / "rst"
```

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -188,50 +259,69 @@ def strip_ansi_codes(text: str) -> str:
...
+    sphinx_app_setup: Callable[[Path], SphinxTestApp],
+    monkeypatch: pytest.MonkeyPatch,
 ) -> None:
     ### Test function to check rules in the given rst file
     # The function uses the SphinxTestApp to build the documentation
     # and checks for the expected/unexpected warnings.
-    rst_data = extract_test_data(RST_DIR / rst_file)
-    if not rst_data:
+    rst_data_raw, parsed_checks_raw = extract_test_data(RST_DIR / rst_file)
+    if not rst_data_raw:
```
AlexanderLanin: _AI-assisted review (Claude):_

`RstData` is a plain dataclass with no `__bool__` override, so `not rst_data_raw` is always `False`. `extract_test_data` also never returns `None` — it always constructs and returns an `RstData`. This guard is dead code and should be removed.

```suggestion
    rst_data = group_test_data(rst_data_raw, parsed_checks_raw)
```

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -112,38 +128,97 @@ def parse_line_for_message(line: str) -> str:
...
+                if offset is None:
+                    rst_data.syntax_errors.append(
+                        f"Warning lines have to have a target warning line like `EXPECT[+1]`. Following line does not have this: \n\t{line}"
+                    )
+                    continue
+                # Offset == 1 means that there is no newline between 'EXPECT/-NOT' and the '.. xyz'.
+                # This is not allowed as this will lead to a silent parsing error and the need will not be registered
+                if offset == 1:
+                    rst_data.syntax_errors.append(
+                        "Warning lines have '+1' as offset. There *HAS* to be a new line between Warning Statement and need."
```
AlexanderLanin: _AI-assisted review (Claude):_

Missing space between the two adjacent string literals — the concatenated message reads `"...and need.Please add..."` with no space before "Please".

```suggestion
                        "Warning lines have '+1' as offset. There *HAS* to be a new line between Warning Statement and need. "
                        "Please add a new line and increase the offset accordingly to the following line:\n\t"
```

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -188,50 +259,69 @@ def strip_ansi_codes(text: str) -> str:
...
 
     # Collect the warnings
     raw_warnings = app.warning.getvalue().splitlines()
-    warnings = [strip_ansi_codes(w) for w in raw_warnings if "score_metamodel" in w]
+    warnings = [strip_ansi_codes(w) for w in raw_warnings]
 
     # Enable this if you need to see errors for debugging purposes
     # print(
-    #     "\n".join(strip_ansi_codes(w) for w in raw_warnings if "score_metamodel" in w)
+    #     "\n".join(strip_ansi_codes(w) for w warnings
```
AlexanderLanin: _AI-assisted review (Claude):_

Syntax error in the commented-out debug block: `for w warnings` is missing `in`. Should be `for w in warnings`. Low priority since it is commented out, but if someone uncomments this during debugging it will be a head-scratcher.

```suggestion
    #     "\n".join(strip_ansi_codes(w) for w in warnings
```

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -188,50 +259,69 @@ def strip_ansi_codes(text: str) -> str:
...
+    monkeypatch.chdir(app.srcdir)
 
     # Build the documentation with the enabled checks
     app.config.score_metamodel_checks = rst_data.enabled_checks
     app.build()
 
     # Collect the warnings
     raw_warnings = app.warning.getvalue().splitlines()
-    warnings = [strip_ansi_codes(w) for w in raw_warnings if "score_metamodel" in w]
+    warnings = [strip_ansi_codes(w) for w in raw_warnings]
```
AlexanderLanin, line 302: _AI-assisted review (Claude):_

The `score_metamodel` filter was removed, so `warnings` now contains every Sphinx warning (image missing, unknown reference, etc.), not just those from this extension. The position-based filter in `filter_warnings_by_position` should prevent false positives for EXPECT, but EXPECT-NOT checks could fail spuriously if Sphinx emits an unrelated warning on the same line as the need. Worth a conscious decision: was this removal intentional (e.g. to catch sphinx-needs warnings that were being silently swallowed before)?

If intentional, a short comment explaining the reason here would be helpful. If not, the original filter should be restored.

AlexanderLanin:

**Critical: 0 | Important: 1 | Suggestions: 3**

Overall this is a clear improvement: the new `#EXPECT[+N]` syntax makes test intent explicit, the line-offset validation catches malformed test files before Sphinx runs, and replacing `os.chdir` with `monkeypatch.chdir` is the right call for test isolation. The refactoring into `extract_test_data` + `group_test_data` is clean.

**Important**
- Removing the `score_metamodel` filter on warnings widens the pool to all Sphinx warnings. The position filter protects EXPECT checks, but EXPECT-NOT checks could spuriously fail if Sphinx emits an unrelated warning on the same line as a need. Needs a clear justification (inline comment or PR description). See inline comment on line 295.

**Suggestions**
- `DOCS_DIR` and `logger` are dead code — neither is referenced after this PR. Drop both (line 25, line 32).
- `if not rst_data_raw` is an unreachable branch — `RstData` is always truthy and `extract_test_data` never returns `None` (line 270).
- Missing space in the `offset == 1` error message: `"...and need.Please add..."` (line 166).
- Commented-out debug `print` has a syntax error: `for w warnings` → `for w in warnings` (line 299).

MaximilianSoerenPollak:
@AlexanderLanin  I in cooperated some of the Review comments, have another look if you can.

docs/internals/extensions/rst_filebased_testing.md
```diff
@@ -36,18 +36,37 @@ Message text which is expected/not expected during the
...
 
 This example verifies that the warning message
 *std_wp__test__abcd: is missing required attribute: \`status\`*
 is shown during the Sphinx build. Only the *check_options* check is enabled.
+
+With the '[+2]' after the 'EXPECT' we tell the parser that we want this warning
+to be emitted and checked for 2 lines underneath
+
+There is multiple things that are not allowed for example, you need to have
+a new line between the EXPECT/-NOT and the need that it refers to
```
a-zw, line 62: Why is that?

MaximilianSoerenPollak: There needs to be a new line between the `EXPECT/-NOT` and the need. 
If that is *not* the case, then our parser sees it correctly (as we just filter for '.. ' and '::') but sphinx-needs or more precisely docutils thinks it is normal text and doesn't actually generate the need itself. 
Therefore we have a miss match of what we see and what sphinx sees. This is not fixable on our side (making the need when sphinx doesn't see it). So we have to enforce (lightly) that there is a new line.

docs/internals/extensions/rst_filebased_testing.md
```diff
@@ -36,18 +36,37 @@ Message text which is expected/not expected during the
                     This message is checked for the Sphinx-Needs directive
                     specified after the EXPECT/EXPECT-NOT statement.
 
+This message needs a '[+x]'offset after the 'EXPECT/-NOT' that should point to the need
+that should (not) emit the warning.
```
a-zw, line 40: It seems to practically always be `[+2]`. So it feels more like unnecessary boilerplate.

MaximilianSoerenPollak: That is currently cause we only have 1 EXPECT there. If we had 2 or more it would  be `
EXPECT[+3]: ...
EXPECT[+2]: ...

etc.  
It truly is only an offset that EXACTLY tells us where they want the warning to be. 
This way we can also safeguard a bit against sphinx not seeing a need but we do.

AlexanderLanin (dismissed):

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -19,16 +19,15 @@
...
 RST_DIR = Path(__file__).absolute().parent / "rst"
-DOCS_DIR = Path(__file__).absolute().parent.parent.parent.parent.parent
-TOOLING_DIR_NAME = "src"
 
 ### List of relative paths of all rst files in RST_DIR
-
-
 RST_FILES = [str(f.relative_to(RST_DIR)) for f in Path(RST_DIR).rglob("*.rst")]
 
+logger = get_logger(__file__)
```
AlexanderLanin: _AI-assisted review (Claude):_

`logger` is imported and instantiated here but never called anywhere in this file. Dead code — drop both lines.

```suggestion
```

AlexanderLanin (commented):

---

Good shape. The previous round's concerns are all addressed:

- Dead `DOCS_DIR` removed ✅
- `if not rst_data_raw` dead guard removed ✅
- Missing space in the offset-1 error message fixed ✅
- Commented-out debug print syntax error fixed ✅
- Warning filter broadening: the comment on line 302–304 explains the rationale. Acceptable. ✅
- `monkeypatch.chdir` replacing `os.chdir` ✅

One small thing remaining: `logger` (lines 22 + 29) is imported and instantiated but never called. See inline comment.

AlexanderLanin (approved):

PR #565, opened 2026-05-28
test: simplify consumer tests

AlexanderLanin:
## Why

The old consumer test was a single monolithic test function that ran all repos and commands sequentially, using `os.chdir()` to navigate between repo directories. This caused several problems:

- A single test failure blocked all remaining repos and commands — no partial results.
- `os.chdir()` made the working directory a global side-effect, causing subtle failures when commands ran in the wrong directory.
- Warnings were parsed from captured stderr output with ANSI stripping and logger categorization, but only visible in the log file — not in pytest's summary. The first warning terminated the process immediately, hiding all subsequent ones.
- The `--repo` filter option required comma-separated repo names and silently fell back to running all repos on typos, making targeted local runs error-prone.
- Output was written to `consumer_test.log` via `rich.Console`, requiring a separate `cat` step in CI to surface it. The CI workflow had a `summarize` job to merge per-repo XML reports and artifacts.
- `test_commands` was a separate field from `commands`, inconsistently populated — only `module_template` used it, with no tests actually wired up.

## What

**Parametrized test structure** — the single test function is replaced with a `@pytest.mark.parametrize` test over every combination of `(repo, override_type, command)`. Each combination is now an independent test case that can pass, fail, or be xfailed individually. No more sequential blocking.

**No `os.chdir()`** — every subprocess call passes an explicit `cwd=` argument. The working directory never changes.

**`local` and `remote` override types as first-class test dimensions** — instead of running both override types inside a single test, each is a separate parametrized case. Remote tests check at collection time whether the commit is pushed and fail with a clear reason if not.

**Warnings forwarded to pytest** — `WARNING` lines are forwarded via `warnings.warn()` as the process runs; pytest collects them individually in its warnings summary. The process runs to completion before failing.

**Plain stdout instead of log file** — output goes to `print()` directly, visible with `-s`. The `consumer_test.log` file and the CI artifact upload machinery are removed.

**`-k` replaces `--repo`** — repo and override type filtering uses standard pytest `-k` syntax, consistent with the rest of the test suite.

## Changes

- `src/tests/test_consumer.py` — full rewrite: parametrized tests, no `os.chdir`, `local`/`remote` override split, warning forwarding, `rich` dependency removed, `get_current_git_hash` from helper_lib used instead of a local reimplementation
- `src/tests/conftest.py` — `--repo` option removed; `--disable-cache` kept
- `.github/workflows/consumer_test.yml` — matrix updated to `"<repo> and local"` / `"<repo> and remote"` entries passed as `-k` filter; `summarize` job and artifact upload removed; `--lockfile_mode=error` removed from `ide_support` call
- `.gitignore` — added `/.codex` and `/.claude`

### DIFF
--- a/src/tests/test_consumer.py
+++ b/src/tests/test_consumer.py
@@ -10,51 +10,40 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
-import os
+"""
+Consumer tests: verify that downstream repos build successfully against this branch.
+
+Via Python (requires ide_support to have been run first):
+    bazel run //:ide_support  # once, to set up .venv_docs
+    python -m pytest src/tests/test_consumer.py -s
+    python -m pytest src/tests/test_consumer.py -k "score and local"
+    python -m pytest src/tests/test_consumer.py -k "process_description"
+    python -m pytest src/tests/test_consumer.py --disable-cache
+
+Known non-passing tests are sometimes marked xfail and do not count as failures.
+"""
+
 import re
 import shutil
 import subprocess
-from collections import defaultdict
-from dataclasses import dataclass, field
+import warnings
+from dataclasses import dataclass
 from pathlib import Path
-from typing import cast
 
 import pytest
-from _pytest.config import Config
 from pytest import TempPathFactory
-from rich import box, print
-from rich.console import Console
-from rich.table import Table
 
-from src.helper_lib import find_git_root, get_github_base_url
+from src.helper_lib import find_git_root, get_current_git_hash, get_github_base_url
 
-"""
-This script's main usecase is to test consumers of Docs-As-Code with
-the new changes made in PR's.
-This enables us to find new issues and problems we introduce with changes
-that we otherwise would only know much later.
-There are several things to note.
+# ---------------------------------------------------------------------------
+# Logging
+# ---------------------------------------------------------------------------
 
-- The `print` function has been overwritten by the 'rich' package to allow for richer
-text output.
-- The script itself takes quiet a bit of time, roughly 5+ min for a full run.
-- If you need more output, enable it via `-v` or `-vv`
-- Start the script via the following command:
-    - bazel run //:ide_support
-    - .venv_docs/bin/python -m pytest -s src/tests
-    (If you need more verbosity add `-v` or `-vv`)
-"""
-
-# Max width of the printout
-# Trial and error has shown that 80 the best value is for GH CI output
-len_max = 120
 CACHE_DIR = Path.home() / ".cache" / "docs_as_code_consumer_tests"
-log_file_name = "consumer_test.log"
-# Need to ignore the ruff error here. Due to how the script is written,
-# can not use a context manager to open the log file, even though it would be preferable
-# In a future re-write this should be considered.
-log_fp = open(log_file_name, "a", encoding="utf-8")  # noqa: SIM115
-console = Console(file=log_fp, force_terminal=False, width=120, color_system=None)
+
+# ---------------------------------------------------------------------------
+# Data model
+# ---------------------------------------------------------------------------
 
 
 @dataclass
@@ -62,24 +51,6 @@ class ConsumerRepo:
     name: str
     git_url: str
     commands: list[str]
-    test_commands: list[str]
-
-
-@dataclass
-class BuildOutput:
-    returncode: int
-    stdout: str
-    stderr: str
-    warnings: dict[str, list[str]] = field(default_factory=dict)
-
-
-@dataclass
-class Result:
-    repo: str
-    cmd: str
-    local_or_git: str
-    passed: bool
-    reason: str
 
 
 REPOS_TO_TEST: list[ConsumerRepo] = [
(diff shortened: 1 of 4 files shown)


src/tests/test_consumer.py
```diff
@@ -113,614 +82,284 @@ class Result:
...
-
             while i < len(lines) and depth > 0:
                 depth += lines[i].count("(") - lines[i].count(")")
                 i += 1
-
-            # Extract the block
             block = lines[block_start:i]
             block_text = "\n".join(block)
-
-            # Comment out if it's for score_docs_as_code
```
MaximilianSoerenPollak, line 227: I if there is a reason for deleting the comments, they could seem helpful?

src/tests/test_consumer.py
```diff
@@ -113,614 +82,284 @@ class Result:
...
             else:
                 result.extend(block)
         else:
             result.append(line)
             i += 1
+    return "\n".join(result) + ("\n" if content.endswith("\n") else "")
 
-    return "\n".join(result) + ("\n" if module_content.endswith("\n") else "")
 
+_BAZEL_DEP_PATTERN = r"""^bazel_dep\(name = ["']score_docs_as_code["'](?:, version = ["'][^"']+["'])?\)"""
```
MaximilianSoerenPollak, line 132: 😩 Does this actually do what we want? This looks a bit complicated?

MaximilianSoerenPollak: Maybe a comment on what it does exactly?

src/tests/test_consumer.py
```diff
@@ -10,76 +10,47 @@
 #
```
MaximilianSoerenPollak, line 10: I see you deleted all of the logging / output generation etc. 

Are you very sure this works? 
The problem with the old one was that the sphinx output generated *too much* output and therefore broke the pipe and the Ci/CD workflow of the consumer tests. 

If you just, as I read it here, output the stdout from the command, this might be the same issue, this will break again then and we have no idea why (if we delte the `-vv` in the workflow call). 


Is this correct or am I missing something here?

MaximilianSoerenPollak (changes requested):
One important question, (regaridng logging deletion) 
Other than that most things look sensible

src/tests/test_consumer.py
AlexanderLanin: Same regex as before... just replaced `"` with `["']` to allow both styles. Will add a comment.

src/tests/test_consumer.py
AlexanderLanin: I have no idea how this is connected, I deleted it after observing that there are no problems at all. The tests run with -vv and dont produce "too much" output. Even with extra fake warnings etc, its all fine.

MaximilianSoerenPollak: The Pipe operation has a limit of 64kb otherwise it fills up and breaks. 
We used to overflow it with the old style, though now maybe with pytest this isn't an issue now as we use that for the 'output' only, not logging stuff.

src/tests/test_consumer.py
MaximilianSoerenPollak: Ohh, Alright, I'm not the smartest in the buisness.

src/tests/test_consumer.py
AlexanderLanin: Yeah, seems to be a non issue

src/tests/test_consumer.py
```diff
@@ -10,76 +10,47 @@
...
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
-import os
+"""
+Consumer tests: verify that downstream repos build successfully against this branch.
+
+Via Python (requires ide_support to have been run first):
```
MaximilianSoerenPollak, line 16: Nit: (can be done some other time) 

If this is truly a requirement we could just do an assert that `.venv_docs` exists and otherwise exit early.

src/tests/test_consumer.py
```diff
@@ -113,614 +82,292 @@ class Result:
...
-        # Level 0-2: Capture output and parse later
-        out = subprocess.run(cmd.split(), capture_output=True, text=True)
-        BR = BuildOutput(
-            returncode=out.returncode,
-            stdout=str(out.stdout),
-            stderr=str(out.stderr),
-        )
+    assert process.stdout is not None
+    for line in process.stdout:
+        print(line, end="")
```
MaximilianSoerenPollak, line 232: Is this on purpose or left over from debugging?

MaximilianSoerenPollak: Can be removed in future PR's if that is the case.

AlexanderLanin: No that's for pytest with -s mode. As it's run from workflows as well.

MaximilianSoerenPollak (approved):
For me this looks fine. 

The two points I commented on are minor and can be removed in a future PR. 

With this change we might can look at testing / expanding the tests soon.

src/tests/test_consumer.py
AlexanderLanin: We can add bazel support as well, but thats a separate PR. This was like that forever.

PR #568, opened 2026-05-29
perf: Split metamodel test

a-zw:
Splitting up the slowest test for better caching. Also, removing needs.json downloads.

## 🚨 Impact Analysis

## ✅ Checklist


src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -287,6 +287,20 @@ def test_rst_files(
...
+    if local_needs.exists():
+        shutil.copy(local_needs, app.srcdir)
+        app.config.needs_external_needs = [
+            {k: v for k, v in ext.items() if k != "json_url"} | {"json_path": "needs.json"}
+            if "json_url" in ext
+            else ext
+            for ext in (app.config.needs_external_needs or [])
+        ]
+    else:
+        app.config.needs_external_needs = []
```
MaximilianSoerenPollak: Is this needed, now that you have changed the conf.py anyway?

src/extensions/score_metamodel/BUILD
```diff
@@ -74,16 +74,46 @@ score_pytest(
...
-        ],
-    ) + ["tests/rst/conf.py"],
+    data = glob(["tests/rst/architecture/*.rst"]) + ["tests/rst/conf.py"],
+    pytest_config = "//:pyproject.toml",
+    deps = [":score_metamodel"],
+)
+
+score_pytest(
+    name = "file_based_tests_attributes",
+    size = "medium",
```
MaximilianSoerenPollak: Are these all medium? 

I don't even know how we determin size, don''t think it matters atm tbh.

src/extensions/score_metamodel/BUILD
```diff
@@ -74,16 +74,46 @@ score_pytest(
...
 score_pytest(
-    name = "file_based_tests",
+    name = "file_based_tests_architecture",
     size = "medium",
     srcs = ["tests/test_rules_file_based.py"],
-    # All requirements already in the library so no need to have it double
-    data = glob(
-        [
-            "tests/**/*.rst",
-            "tests/**/*.yaml",
```
MaximilianSoerenPollak, line 84: Do not forget to also add the .yaml file, one tests needs that.

MaximilianSoerenPollak (changes requested):
Only 2 comments. 
The size thing is irrelevant for this PR and something we need to address overall

src/extensions/score_metamodel/tests/test_rules_file_based.py
a-zw: It is removed in general, but added only for file_based_tests_attributes. This one needs it but the other three do not.

src/extensions/score_metamodel/BUILD
a-zw: The unit tests are declared as "small" already, but yeah I don't understand the point of this attribute either. Removing it ...

src/extensions/score_metamodel/BUILD
a-zw: The tests are all passing. Maybe you think of `test_metamodel_load.py`? That one is part of `:unit_tests`.

MaximilianSoerenPollak: Ah yes, I did.

src/extensions/score_metamodel/tests/test_rules_file_based.py
MaximilianSoerenPollak: Would it then not be easier to just add it anyway in conf.py? As now it's local and doesn't have to be downloaded? 
We could then save all this logic here. 

Or do you think it makes sense to have this piece of code to guarantee only the one thing that needs it, has it?

a-zw: Good point. After trying, I think you are right.

The point is to avoid downloading the file. Only using it in certain tests does not really help anybody and "costs" this added logic. So I changed it.

MaximilianSoerenPollak (approved):
Looks good, this should speed things up and give more clarity 💯

PR #572, opened 2026-06-03
Wip tmp tracability improvements

MaximilianSoerenPollak:
# ⚠️ This is a TMP Branch, just made to see diff's better. DO NOT MERGE ⚠️ 

## Here is a waffle as thanks 🧇


docs/conf.py
```diff
@@ -18,3 +19,4 @@
 extensions = [
     "score_sphinx_bundle",
 ]
+matplotlib.rcParamsDefault["savefig.bbox"] = "tight"
```
MaximilianSoerenPollak, line 22: Probably should have this in the extension so it's set for all things that import docs-as-code

src/extensions/score_metamodel/checks/standards.py
```diff
@@ -18,7 +18,7 @@
```
MaximilianSoerenPollak, line 18: Ignore this file, has not yet been changed / updated to new architecture

src/extensions/score_metamodel/__init__.py
```diff
@@ -31,9 +31,6 @@
     ProhibitedWordCheck as ProhibitedWordCheck,
```
MaximilianSoerenPollak, line 29: Was moved to it's own extension

src/extensions/score_metrics/tests/test_traceability_dashboard.py
```diff
@@ -0,0 +1,218 @@
+# *******************************************************************************
```
MaximilianSoerenPollak: Tests commented out (to test the metrics stuff). 
Will be redone once arch is finished

src/extensions/score_metrics/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+def setup(app: Sphinx) -> dict[str, str | bool]:
+    app.add_config_value(
+        "score_metamodel_requirement_types",
+        "",
+        rebuild="env",
+        description=(
+            "Comma-separated list of need types treated as requirements for "
+            "traceability metrics. If empty, requirement types are autodiscovered "
+            "from needs_types tags (requirement, requirement_excl_process)."
```
MaximilianSoerenPollak, line 61: Should be moved to other function. `calculate_need_metrics`

src/extensions/score_metrics/sphinx_filters.py
```diff
@@ -55,14 +55,15 @@ def _matches_source_selector(need: NeedItem, selector: str) -> bool:
 
 
 def generic_pie_linked_items(
-    needs: list[NeedItem], results: list[int], **kwargs: str | int | float
+    needs: list[NeedItem], results: list[int], **kwargs: dict[str, str | int | float]
```
MaximilianSoerenPollak: Kwargs typhint should be 'Any' maybe?

src/extensions/score_metrics/sphinx_filters.py
```diff
@@ -164,3 +165,11 @@ def generic_pie_items_in_relationships(
...
     results.append(referenced_once)
     results.append(referenced_multiple)
+
+
+def get_metrics_from_generated_json(
+    needs: list[NeedItem], results: list[int], **kwargs: str | int | float
+) -> None:
+    metrics_json_path = str(kwargs.get("arg1", "_build/metrics.json"))
+    # TODO: Read the Metrics.json and give back the data
+    results
```
MaximilianSoerenPollak: Delete

src/extensions/score_metrics/traceability_dashboard.py
```diff
@@ -0,0 +1,256 @@
...
+def get_just_metrics(needs: list[Any], results: list[int], **kwargs: Any) -> None:
+    """Append selected metric values without any total/remainder calculation.
+
+    All kwarg values are interpreted as colon-separated metric paths and
+    appended to ``results`` in insertion order.
+    """
+    _get_key_values(results, [str(value) for value in kwargs.values()])
+
+
+# def set_default_include_external(include_external: bool) -> None:
```
MaximilianSoerenPollak: Re-Read commented out functions, see what 'general' functions you are missing from there if any.

src/extensions/score_metrics/traceability_metrics.py
```diff
@@ -0,0 +1,208 @@
+# *******************************************************************************
```
MaximilianSoerenPollak, line 1: Many of this has been copied from the old '_write_metrics' function. 
But simplified.

src/extensions/score_source_code_linker/xml_parser.py
```diff
@@ -219,9 +219,10 @@ def read_test_xml_file(file: Path) -> tuple[list[DataOfTestCase], list[str], lis
             properties_element = testcase.find("properties")
```
MaximilianSoerenPollak, line 219: Should this be in a seperate PR?

MaximilianSoerenPollak (commented):
Commenting some things.

docs/conf.py
MaximilianSoerenPollak: @AlexanderLanin thoughts on this?

src/extensions/score_metrics/__init__.py
MaximilianSoerenPollak: Actually this should stay. The 'setup''  function should add everything the whole extension needs.

src/extensions/score_metrics/traceability_dashboard.py
MaximilianSoerenPollak: Everything from the commented out functions is easily doable in needpie / needlist.

docs/how-to/dashboards_and_quality_gates.rst
```diff
@@ -40,20 +40,35 @@ Typical Setup
```
MaximilianSoerenPollak, line 40: Docs has been generated by Copilot. 

Has not yet been updated / looked through for the new changes made.

AlexanderLanin (commented):

**Critical: 0 | Important: 7 | Suggestion: 3**

This is a useful architectural direction — extracting metrics into a dedicated `score_metrics` extension makes sense. But several issues need to be addressed before this is ready to merge.

---

### Bugs that will cause silent failures at runtime

The two most dangerous are easy to miss because they produce zero-error output rather than a crash:

1. **Typo `"reqiurement"` in autodiscovery** (`score_metrics/traceability_metrics.py:141`) — when `score_metamodel_requirement_types` is not set, `get_need_types_by_tags` is called with `{"reqiurement"}` instead of `{"requirement"}`. This returns an empty list silently, producing zero-coverage metrics with no warning to the user.

2. **`CALCULATED_METRICS` always `{}` at write time** (`score_metrics/__init__.py:22`) — `calculate_full_need_metrics` is imported from `score_metrics.traceability_metrics`, but `CALCULATED_METRICS` is imported from `src.extensions.score_metrics.traceability_metrics`. These are two different module objects. After `calculate_full_need_metrics` runs and sets `traceability_metrics.CALCULATED_METRICS = output`, the `CALCULATED_METRICS` name in `__init__` still points to the original empty dict. The metrics JSON file will always be written as `{}`.

---

### Important issues

**`score_sphinx_bundle/__init__.py` — `myst_heading_anchors` and `myst_heading_slug_func` removed**

These two settings were added specifically to fix cross-reference heading slug mismatches. Removing them reinstates a broken behavior that the settings fixed. This looks unintentional.

**`score_metrics/traceability_metrics.py:105–107` — `None + list` TypeError**

```python
partially: list[str] = test_need.get("partially_verifies")
fully: list[str] = test_need.get("fully_verifies")
refs = set(partially + fully)
```

`dict.get()` returns `None` when the key is absent. `None + list` raises `TypeError`. Needs `or []` guards.

**`filter_is_external(include_external)` semantics** (`traceability_metrics.py:159–160`)

`NeedsView.filter_is_external(value)` keeps needs where `is_external == value`. So `filter_is_external(True)` returns *only* external needs, not "all needs including external". When `include_external=True`, local requirements would be excluded — the opposite of the intent.

**`check_needs_extends.py` deleted with no migration**

The `needextend` enforcement checks (no replace of already-set fields, no delete, no string append) are silently gone. If this is intentional, it needs a note explaining why the enforcement is no longer needed.

**`test_rules_file_based.py` is a regression**

The PR's version uses `#EXPECT:` syntax. All current RST test files use `#EXPECT[+N]:` with offset notation. The PR's parser won't match any of them — it will silently find zero expectations and pass every test file vacuously. This is the worst outcome: CI stays green while all file-based checks are disabled.

**`score_metamodel/checks/standards.py` — circular dependency**

`standards.py` now imports from `score_metrics.sphinx_filters`, and `score_metrics/traceability_metrics.py` imports `ScoreNeedType` from `score_metamodel`. This creates a `score_metamodel → score_metrics → score_metamodel` cycle. It's fragile at Python import time and is an undeclared Bazel dependency.

---

### Suggestions

**`sphinx_filters.py:210` — stray `print(results)`** will print to stdout on every Sphinx build.

**`_write_metrics_json` should check the exception parameter** — `build-finished` passes the build exception as the second argument. Writing `metrics.json` unconditionally on a failed build may write partial/empty data.

**`score_metamodel.sphinx_filters` has no compat shim** — consumers using `:filter-func: score_metamodel.sphinx_filters.generic_pie_linked_items` will get `ModuleNotFoundError` after the move.

---

Overall: the split is the right direction. The typo, the mixed import paths, the `myst_heading` removal, and the test harness regression are the most urgent to fix.

MaximilianSoerenPollak:
Current Test Coverage: 
image

MaximilianSoerenPollak:
#578 will merge this branch correctly.

PR #578, opened 2026-06-05
Wip tmp tracability improvements

MaximilianSoerenPollak:
## 📌 Description
Overall improvements of the structure. 
It now is it's own Sphinx Extension with all benefits from that. 
Simplified a lot of the code to enhance performance and readability.

## 🚨 Impact Analysis

## ✅ Checklist


AlexanderLanin (commented):

**Critical: 2, Important: 5, Suggestions: 4**

This PR does good structural work — extracting metrics into a dedicated `score_metrics` extension is the right call, and the `Label("//...")` fix in `docs.bzl` is a genuine correctness improvement for cross-repo macro usage. However there are two critical issues that will cause immediate CI/build failures, plus several important gaps in error handling and test coverage.

**Critical issues to fix before merge:**
1. `actions/checkout@v6.0.2` does not exist — the workflow will fail at checkout
2. Broken import `from src.extensions.score_metamodel.tests import need` in `test_traceability_metrics_json_generation.py` — collection will fail

**Other important issues:**
- `_write_metrics_json` writes metrics even when `exception` is not None — CI gate will read partial/empty JSON and either pass trivially or crash
- `is_external` warning in `check_needs_extends.py` does not `continue` — the extend is still applied after warning
- `test_traceability_metrics.py` is entirely commented out, leaving `calculate_full_need_metrics` untested
- Remote matrix legs call `pytest.fail()` instead of `pytest.skip()` when remote is unavailable

Inline comments below for specifics.

src/extensions/score_metrics/tests/test_traceability_metrics_json_generation.py
```diff
@@ -0,0 +1,233 @@
...
+import score_metrics.traceability_metrics as metrics
+from score_metamodel import ScoreNeedType
+from sphinx_needs.data import NeedsView
+from sphinx_needs.need_item import NeedItem
+
+from score_pytest.attribute_plugin import add_test_properties
+
+# from score_metamodel.tests import need as test_need
+from src.extensions.score_metamodel.tests import need as test_need
+
```
AlexanderLanin: _AI-assisted review (Claude):_ **Critical — broken import, test collection will fail**

The `pythonpath` in `pyproject.toml` is set to `["src/extensions/"]`, so `score_metamodel` is importable as a top-level package. The `src.extensions.` prefix is not resolvable from there. This will cause an `ImportError` at collection time, breaking the entire `score_metrics` test target.

The correct import is already on the line above (commented out):

```suggestion
from score_metamodel.tests import need as test_need
```

src/extensions/score_metrics/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _write_metrics_json(app: Sphinx, exception: Any | None) -> None:
+    """
+    Write a schema-v1 metrics.json alongside needs.json in the build output.
+    """
+    if exception:
+        logger.error(f"Sphinx-Exception at end of build: {exception}")
+    out_path = Path(app.outdir) / "metrics.json"
+    out_path.parent.mkdir(parents=True, exist_ok=True)
+    out_path.write_text(json.dumps(CALCULATED_METRICS, indent=2), encoding="utf-8")
```
AlexanderLanin, line 49: _AI-assisted review (Claude):_ **Important — writes partial/empty metrics even when the build failed**

When `exception` is not `None`, the handler logs the error but then continues to write `CALCULATED_METRICS` (which may be `{}` or stale) to disk. The CI gate will read this file and either pass trivially (empty denominator → 100%) or crash with a `KeyError`.

Add an early return:

```suggestion
    if exception:
        logger.error(f"Sphinx-Exception at end of build: {exception}")
        return
```

src/extensions/score_metrics/tests/test_traceability_metrics.py
```diff
@@ -0,0 +1,201 @@
...
+            "type": "testcase",
+            "partially_verifies": "LOCAL_REQ",
+            "fully_verifies": "",
+            "is_external": False,
+        },
+    ]
+
+
+# def test_filter_requirements_defaults_to_local_only() -> None:
+#     filtered = filter_requirements(
```
AlexanderLanin: _AI-assisted review (Claude):_ **Important — entire test file is commented out, zero tests run**

Every test function in this file is inside a block comment. The file contributes zero coverage. `calculate_full_need_metrics` (the function that aggregates per-type metrics and populates `CALCULATED_METRICS`) has no tests at all in the new codebase.

Either delete this file entirely and add a test for `calculate_full_need_metrics`, or rewrite the tests against the current API. A test file with all tests commented out is worse than no file — it signals intent without providing coverage.

src/extensions/score_metrics/traceability_metrics.py
```diff
@@ -0,0 +1,195 @@
...
+        "schema_version": "1",
+        "generated_by": "sphinx_build",
+        "overall_metrics": overall_metrics,
+        "metrics_by_type": metrics_by_type,
+        "tests": test_metrics,
+    }
+    # Save the metrics in a Global Variable to enable access from other parts.
+    # Not a great solution but it is needed, as needpie filter functions for example
+    # can not access 'app'.
+    CALCULATED_METRICS.update(output)
```
AlexanderLanin, line 195: _AI-assisted review (Claude):_ **Important — global state not cleared between incremental builds**

`CALCULATED_METRICS.update(output)` appends to the existing global dict. On an incremental Sphinx rebuild, stale type entries from the previous run survive if the function raises before this line, or if a need type disappears between runs.

```suggestion
    CALCULATED_METRICS.clear()
    CALCULATED_METRICS.update(output)
```

src/extensions/score_metamodel/__init__.py
```diff
@@ -411,7 +270,7 @@ def setup(app: Sphinx) -> dict[str, str | bool]:
     # sphinx-collections runs on default prio 500.
     # We need to populate the sphinx-collections config before that happens.
     # --> 499
```
AlexanderLanin: _AI-assisted review (Claude):_ **Suggestion — stale comment**

The comment says `--> 499` but the code uses `priority=450`.

```suggestion
    # --> 450
```

src/extensions/score_metrics/tests/test_traceability_metrics_json_generation.py
MaximilianSoerenPollak: Tests pass and no import error happens so idk. 
But I can correct it.

src/extensions/score_metrics/__init__.py
MaximilianSoerenPollak: This is wanted behaviour. We still want to have the metrics json in ANY case, as other things might depend on it. 
we do not know what we want to do with the json afterwards, so guranteeing that it is there makes sense in my mind.

src/extensions/score_metrics/tests/test_traceability_metrics.py
MaximilianSoerenPollak: Deleted the non needed file.

src/extensions/score_metamodel/__init__.py
MaximilianSoerenPollak: Adapted comment.

AlexanderLanin (approved):

PR #584, opened 2026-06-08
Docs: Improve XML documentation and logging

### LINKED ISSUE #573: request: improve test xml handling
Including test.xml is rather poorly documented with very bad or no runtime feedback in case it doesnt work.

Examples:
* docu lacks example of real xml with all attributes and properties. Attributes are not documented.
* xml needs to be in test-reports or bazel-testlogs -> not documented & arbitrary
* however bazel-testlogs must exist either way
* no feedback if xml is invalid in some way
* no feedback whether xml was found by docs-a-scode

MaximilianSoerenPollak:
## 📌 Description
The current XML parser part of source_code_linker was not very well documented and some feedback was missing. 
This PR addresses both.

Fixes: #573 

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs/how-to/test_to_doc_links.rst
+++ b/docs/how-to/test_to_doc_links.rst
@@ -28,9 +28,9 @@ The ``score_source_code_linker`` extension parses ``test.xml`` files, extracts t
 (name, file, line, result, and verification properties),
 and creates backlinks on the referenced requirements.
 
-The extension will look for ``test.xml`` files in both ``bazel-testlogs/`` 
-as well as a folder named ``tests-report/``. This folder can be created manually if the tests 
-require some pre-run step or are matrix tests or similar. 
+The extension will look for ``test.xml`` files in both ``bazel-testlogs/``
+as well as a folder named ``tests-report/``. This folder can be created manually if the tests
+require some pre-run step or are matrix tests or similar.
 
 
 Required Properties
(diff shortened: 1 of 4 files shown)


AlexanderLanin (commented):

**Critical: 0 | Important: 2 | Suggestions: 2**

The PR title says "documentation and feedback fixes" but the xml_parser change is a **behavioral change**: tests that were previously silently skipped (no properties / missing properties) are now registered as sphinx needs. That's worth calling out explicitly.

### Important – logger.info called with multiple positional strings

`find_xml_files` passes three separate string arguments to `logger.info`. Python's logging treats extra positional args as printf-style format values — `msg % args`. Since the first string has no `%` format specifiers, this either raises `TypeError: not all arguments converted during string formatting` (printed to stderr and swallowed) or simply drops the extra strings, producing a truncated message. The informative parts of the message will never appear in logs.

Fix: use adjacent string literal concatenation (no commas):
```python
logger.info(
    "Did not find any test.xml files. "
    "If you expected xml files to be found, please ensure that you have ran your tests "
    "and put all testfiles either in tests-reports or bazel-testlogs."
)
```

### Important – incomplete testcases silently become linked needs

Previously, tests with no `<properties>` element or with missing required properties were excluded from `test_case_needs` (early `continue`). Now they are always appended. In `construct_and_add_need`, `file` defaults to `"<placeholder_file>"` and `FullyVerifies`/`PartiallyVerifies` will be empty — so these needs appear in the docs as testcases that verify nothing, with a placeholder file URL.

For test cases where `properties_element is not None` but `is_valid()` returns False, the need is still added (just logged). The test update only checks `isinstance(tcneed, DataOfTestCase)` — it doesn't assert that `tn.name is not None` before reaching `construct_and_add_need`, where `assert tn.name is not None` would crash if a `<testcase>` element lacks a `name` attribute.

This behavioral change is real and intentional, but it needs:
1. A test asserting the crash path is safe (name always present even when properties are absent).
2. Docs updating the "Limitations" list to reflect the new behavior (the removed bullet "All mandatory properties must be present; partial metadata causes the test to be silently skipped" was accurate before — now it's gone but the new behavior description in the How-To is only partial).

### Suggestion – doc grammar

`test_to_doc_links.rst`:
- "How should a test.xml look like?" → "What should a test.xml look like?"
- "In python tests a docstring is taking the place of the description attribute" → "In Python tests, a docstring takes the place of the description attribute."

### Suggestion – Rust section factual change

The previous text described a concrete Rust mechanism (`#[record_property]` from the `test_properties` crate). The replacement ("Currently there is no provided official way to do this in Rust. We are working with the rust community to figure this out.") is a significant factual rollback. If this reflects actual project state, it should be — but worth confirming this was intentional and not a revert of still-valid documentation.

AlexanderLanin (approved):

PR #590, opened 2026-06-10
feat: remove outdated options

AlexanderLanin:
Remove ancient options for sourcelinks and testlinks.

Added new check which produces following warnings:
```
ERROR: while setting up extension score_metamodel: Directive 'stkh_req' has optional_options that overlap with global base options: {'testlink'}.
ERROR: while setting up extension score_metamodel: Directive 'feat_req' has optional_options that overlap with global base options: {'testlink'}.
ERROR: while setting up extension score_metamodel: Directive 'comp_req' has optional_options that overlap with global base options: {'testlink'}.
ERROR: while setting up extension score_metamodel: Directive 'aou_req' has optional_options that overlap with global base options: {'testlink'}.
```

### DIFF
--- a/src/extensions/score_metamodel/metamodel.yaml
+++ b/src/extensions/score_metamodel/metamodel.yaml
@@ -262,8 +262,6 @@ needs_types:
       # req-Id: tool_req__docs_common_attr_security
       security: ^(YES|NO)$
     optional_options:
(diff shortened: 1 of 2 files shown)


MaximilianSoerenPollak (approved):
Reviewed in Call together. 💯

PR #597, opened 2026-06-11
fix: consecutive builds are working again

AlexanderLanin:
See comment in code for explanation.


AlexanderLanin (commented):
We can discuss running docs twice in CI independent of this quick fix

AlexanderLanin:
Note: consumer test failing for score is due to an independent error. score main branch is broken since ~8h.

MaximilianSoerenPollak (approved):

PR #598, opened 2026-06-12
feat: drop regex support for links in metamodel

AlexanderLanin:
simplify the code and the metamodel

### DIFF
--- a/src/extensions/score_metamodel/__init__.py
+++ b/src/extensions/score_metamodel/__init__.py
@@ -165,51 +165,52 @@ def _resolve_linkable_types(
     link_name: str,
     link_value: str,
     current_need_type: ScoreNeedType,
-    needs_types: list[ScoreNeedType],
-) -> list[ScoreNeedType | str]:
-    needs_types_dict = {nt["directive"]: nt for nt in needs_types}
+    needs_types: dict[str, ScoreNeedType],
+) -> list[ScoreNeedType]:
+    # Anything is allowed if the value is "ANY". This allows to bypass the link type validation for specific links.
+    if link_value == "ANY":
+        return list(needs_types.values())
+
     link_values = [v.strip() for v in link_value.split(",")]
-    linkable_types: list[ScoreNeedType | str] = []
+    linkable_types: list[ScoreNeedType] = []
     for v in link_values:
-        if v.startswith("^"):
-            linkable_types.append(v)  # keep regex as-is
+        target_need_type = needs_types.get(v)
+        if target_need_type is None:
+            logger.error(
+                f"In metamodel.yaml: {current_need_type['directive']}, "
+                f"link '{link_name}' references unknown type '{v}'."
+            )
         else:
-            target_need_type = needs_types_dict.get(v)
-            if target_need_type is None:
-                logger.error(
-                    f"In metamodel.yaml: {current_need_type['directive']}, "
-                    f"link '{link_name}' references unknown type '{v}'."
-                )
-            else:
-                linkable_types.append(target_need_type)
+            linkable_types.append(target_need_type)
     return linkable_types
 
 
 def postprocess_need_links(needs_types_list: list[ScoreNeedType]):
     """Convert link option strings into lists of target need types.
 
-    If a link value starts with '^' it is treated as a regex and left
-    unchanged. Otherwise it is a comma-separated list of type names which
-    are resolved to the corresponding ScoreNeedTypes.
+    Parses comma-separated list of type names which are resolved to the corresponding
+    ScoreNeedTypes.
     """
-    for need_type in needs_types_list:
-        try:
-            link_dicts = (
-                need_type["mandatory_links"],
-                need_type["optional_links"],
+    # "mandatory_links_str" is used to keep only metamodel sourced types. That key is so
+    # specific, that noone but our metamodel should be using it.
+    all_need_types = {
+        nt["directive"]: nt for nt in needs_types_list if "mandatory_links_str" in nt
+    }
+
+    for need_type in all_need_types.values():
+        need_type["mandatory_links"] = {
+            link_name: _resolve_linkable_types(
+                link_name, link_value, need_type, all_need_types
             )
-        except KeyError:
-            # TODO: remove the Sphinx-Needs defaults from our metamodel
-            # Example: {'directive': 'issue', 'title': 'Issue', 'prefix': 'IS_'}
-            continue
+            for link_name, link_value in need_type["mandatory_links_str"].items()
+        }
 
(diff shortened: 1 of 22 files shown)


src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -134,19 +133,23 @@ def _validate(attributes_to_allowed_values: dict[str, str], mandatory: bool):
...
-      and returned unchanged.
-    - A ScoreNeedType dict uses its ``mandatory_options.id`` pattern.
+    Note: the pattern is already stored in "id" attribute, this is just a helper
+    function to retrieve it safely.
     """
-    if isinstance(value, str):
-        if value.startswith("^"):
-            return value
-        return f"^{value}__"
+    assert isinstance(value, dict), f"Expected dict for ScoreNeedType, got {value}"
```
MaximilianSoerenPollak, line 143: Can you not do `is instance` to make sure it's actually a ScoreNeedType?

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -160,10 +163,12 @@ def validate_links(
     """
 
     def _validate(
-        attributes_to_allowed_values: AllowedLinksType,
+        attributes_to_allowed_values: dict[str, list[ScoreNeedType]] | None,
         mandatory: bool,
         treat_as_info: bool = False,
     ):
+        assert attributes_to_allowed_values is not None
```
MaximilianSoerenPollak, line 170: Kinda an antipattern no? 
If the typehint is allowed to be None, but here you assert that it should never happen?

src/extensions/score_metamodel/checks/graph_checks.py
```diff
@@ -40,6 +45,7 @@ def eval_need_check(need: NeedItem, check: str, log: CheckLogger) -> bool:
         "<": operator.lt,
         ">=": operator.ge,
         "<=": operator.le,
+        "contains": lambda a, b: b in a if isinstance(a, str) else False,
```
MaximilianSoerenPollak, line 48: This then allows for b to be a non string type. Is that wanted behaviour?

src/extensions/score_metamodel/tests/rst/graph/test_workproduct_aspice_40.rst
```diff
@@ -0,0 +1,65 @@
...
+
+#CHECK: check_metamodel_graph
+
+--- Setup
+
+.. an ASPICE 40 IIC standard requirement (valid link target).
+
+.. std_req:: ASPICE 40 IIC requirement
+   :id: std_req__aspice_40__iic_1
+
```
MaximilianSoerenPollak, line 23: Uhh idk if this will work? 

As a 1 line seperation might interpret it as content for the need. 
Though unsure about that, has been a while.

src/extensions/score_metamodel/__init__.py
```diff
@@ -165,51 +165,52 @@ def _resolve_linkable_types(
...
+    ScoreNeedTypes.
     """
-    for need_type in needs_types_list:
-        try:
-            link_dicts = (
-                need_type["mandatory_links"],
-                need_type["optional_links"],
+    # "mandatory_links_str" is used to keep only metamodel sourced types. That key is so
+    # specific, that noone but our metamodel should be using it.
+    all_need_types = {
```
MaximilianSoerenPollak, line 196: Is this all_need_types? It reads like a filtered version already.

src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -178,7 +178,8 @@ needs_types:
     mandatory_options:
       status: ^(valid|draft)$
     optional_links:
-      complies: std_wp, ^std_req__aspice_40__iic.*$
+      # FIXME: only ^std_req__aspice_40__iic.*$
```
MaximilianSoerenPollak: FIXME Still needed? THis is now implemented with the graph check no?

src/extensions/score_metamodel/metamodel_types.py
```diff
@@ -27,16 +27,17 @@ class ProhibitedWordCheck:
...
     optional_options: dict[str, str]
 
-    mandatory_links: AllowedLinksType
-    optional_links: AllowedLinksType
+    mandatory_links_str: dict[str, str]
+    optional_links_str: dict[str, str]
+
+    # At parse time these are None, as they need to be filled post-parsing when all types are available to resolve the link targets.
+    # The dict maps the link name to a list of need types that are allowed as targets for that link.
+    mandatory_links: None | dict[str, list[ScoreNeedType]]
```
MaximilianSoerenPollak, line 42: Default them to `None` ? Or is that automatic?

MaximilianSoerenPollak (commented):
Some questions

src/extensions/score_metamodel/checks/check_options.py
AlexanderLanin: No, isinstance does not work with TypedDict. As the type info is not a real type AFAIK.

src/extensions/score_metamodel/checks/check_options.py
AlexanderLanin: yeah... its not pretty. This avoid checking the argument on all call sites, which would be even uglier.

src/extensions/score_metamodel/checks/graph_checks.py
AlexanderLanin: no, but I did not figure out how to say "else fail horribly" - without changing how the function is designed.

src/extensions/score_metamodel/tests/rst/graph/test_workproduct_aspice_40.rst
AlexanderLanin: doesnt matter, as next line is a comment. Whether its a comment, real content, or need content. dont care.

src/extensions/score_metamodel/__init__.py
AlexanderLanin: it's "all our types" aka "all types from metamodel"

src/extensions/score_metamodel/metamodel.yaml
AlexanderLanin: changed from fixme to simple comment referring to graph check

src/extensions/score_metamodel/metamodel_types.py
AlexanderLanin: Unfortunately there is no default in TypedDict. All call sites adapted.

MaximilianSoerenPollak (approved):

PR #599, opened 2026-06-15
feat: treat version as int

AlexanderLanin:
Warning: this is based on unreleased process_description

Note: consumer test on score module fails, as that pulls last released process description.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -70,6 +70,11 @@ http_file(
 )
 
 bazel_dep(name = "score_process", version = "1.5.4")
+git_override(
+    module_name = "score_process",
+    commit = "1c7867b85a1c1ca915bd516cb60ef2e9c1904975",
+    remote = "https://github.com/eclipse-score/process_description.git",
+)
 
 # Provide the tools from the devcontainer to Bazel
 bazel_dep(name = "score_devcontainer", version = "1.7.0")
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -72,7 +72,7 @@ http_file(
 bazel_dep(name = "score_process", version = "1.5.4")
 git_override(
     module_name = "score_process",
-    commit = "1c7867b85a1c1ca915bd516cb60ef2e9c1904975",
+    commit = "0113471d1dbb58d0890dae2569b18af69cd2025f",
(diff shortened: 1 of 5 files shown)


src/extensions/score_metamodel/checks/check_options.py
MaximilianSoerenPollak, line 39: Could you not do: 


```suggestion
    if isinstance(raw_value, str | int):
```
to check both?

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -43,9 +43,15 @@ def _get_normalized(need: NeedItem, key: str) -> list[str]:
         raw_list = cast(list[object], raw_value)
         for item in raw_list:
             if not isinstance(item, str):
```
MaximilianSoerenPollak:
```suggestion
if not all(isinstance(item, str) for item in raw_list)
```

Making it shorter / more efficent with list comprehension?

MODULE.bazel
```diff
@@ -70,6 +70,11 @@ http_file(
 )
 
 bazel_dep(name = "score_process", version = "1.5.4")
+git_override(
+    module_name = "score_process",
+    commit = "1c7867b85a1c1ca915bd516cb60ef2e9c1904975",
```
MaximilianSoerenPollak: Probably should now adapt this to the latest commit that was just merged.

MaximilianSoerenPollak (dismissed):
Just some small simplification. 

But overall this looks good and we can merge it. 

All things mentioned are nitpicks and can be done in future PR's

src/extensions/score_metamodel/checks/check_options.py
AlexanderLanin: yes and no. I would keep it separate as casting int to a str feels like a dirty temporary workaround. Hence the comment.

src/extensions/score_metamodel/checks/check_options.py
AlexanderLanin: done

MODULE.bazel
AlexanderLanin: used latest commit hash. requested process 1.6.0 for later today

MaximilianSoerenPollak (approved):
LGTM 💯

PR #604, opened 2026-06-16
feat: `feature_in_id` for single module repositories

### LINKED ISSUE #596: Create PoC to showcase 'feature in id' functionality
### What

We need to re-work the `feature-in-id` check. As with the new module_template repository structure this will always fail. 
So we need a way to have a list or something of acceptable features, components, modules, and only those can be used in the ID's in this project. 

### Acceptance Criteria (DoD)

Have the test work with current module_template folder structure and correctly throw an error if an id does not have the feature, component or module in it.

### How

Have to figure it out

MaximilianSoerenPollak:
## 📌 Description
This is a suggestion/PoC on how we could allow for the `feature_in_id` check to be kept but adapt it so that single feature /module repositories also pass this check. 
This is done here by adding an additional config parameter to the conf.py that defines strings that are also allowed in the feature_part of the id. 
Now only if NONE of the 3 things (docpath, abbreviation or configured param) are found an error is thrown. 

Fixes #596 

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs/how-to/write_docs.rst
+++ b/docs/how-to/write_docs.rst
@@ -92,7 +92,7 @@ The feature part is validated by checking that at least one of the following is 
    :widths: 45 10 45
 
    * - ID
-     - 
+     -
      - Reason
    * - ``feat_saf__fmea__late_message``
      - ✅
--- a/docs/how-to/write_docs.rst
+++ b/docs/how-to/write_docs.rst
@@ -70,3 +70,47 @@ For further documentation on needextends please `look here <https://sphinx-needs
 
    In the future we will enable a check that needextends will only modify needs in the current document.
    You can ensure this by adding `c.this_doc()` to the filter string of the need.
+
+
+Requirement ID Feature Part
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Requirement IDs with 3 parts (defined by the Metamodel) follow the format ``<Type>__<feature>__<Title>``.
+The ``<feature>`` part must relate to where the requirement lives in the documentation,
+so that IDs stay meaningful and traceable as the project evolves.
+
+The feature part is validated by checking that at least one of the following is true:
+
+* A segment of the feature part (split on ``_`` and ``-``) appears in the document's directory path
+* The initials of the feature part's segments appear in the document's directory path
+* The feature part contains a string explicitly allowed via ``required_in_id`` in ``conf.py``
+
+**Examples** — given a requirement in ``internals/safety/fmea/requirements.rst``:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 45 10 45
+
+   * - ID
+     - 
+     - Reason
+   * - ``feat_saf__fmea__late_message``
+     - ✅
+     - ``fmea`` is in the path
+   * - ``feat_saf__safety_fmea__late_message``
+     - ✅
+     - ``safety`` and ``fmea`` are in the path
+   * - ``feat_saf__sf__late_message``
+     - ✅
+     - ``sf`` are the initials of ``safety_fmea``, which is in the path
+   * - ``feat_saf__blabla__late_message``
+     - ❌
+     - ``blabla`` has no relation to the path ``internals/safety/fmea``
+
+To explicitly allow a feature part that intentionally doesn't match the path
+(e.g. in a single module repository), add a matching string to ``required_in_id`` in ``conf.py``:
+
+.. code-block:: python
+
+   # conf.py
+   required_in_id = ["persistenc"]
--- a/src/extensions/score_metamodel/checks/id_contains_feature.py
+++ b/src/extensions/score_metamodel/checks/id_contains_feature.py
@@ -58,19 +58,36 @@ def id_contains_feature(app: Sphinx, need: NeedItem, log: CheckLogger):
         for featurepart in featureparts
         if featureparts and featurepart and docname
     )
+    allowed_parts_from_config = app.config.required_in_id
+    found_part_from_config = any(
+        part_from_config.lower() in need.get("id")
+        for part_from_config in allowed_parts_from_config
+        if allowed_parts_from_config
+    )
 
     # allow abbreviation of the feature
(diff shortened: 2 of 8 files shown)


src/extensions/score_metamodel/checks/id_contains_feature.py
```diff
@@ -58,19 +58,36 @@ def id_contains_feature(app: Sphinx, need: NeedItem, log: CheckLogger):
         for featurepart in featureparts
         if featureparts and featurepart and docname
     )
+    allowed_parts_from_config = app.config.required_in_id
+    found_part_from_config = any(
+        part_from_config.lower() in need.get("id")
+        for part_from_config in allowed_parts_from_config
```
AlexanderLanin, line 64: _AI-assisted review (Claude):_

The match is against the **full need ID** (e.g. `feat_req__blabla__testing`) rather than the feature part alone (`parts[1]`). This means a config entry like `"feat"` or `"req"` would silently suppress the check for every 3-part ID, regardless of the actual feature value.

The check should compare against `feature` (i.e. `parts[1]`) to match the documented intent: _"strings that are also allowed in the feature\_part of the id"_.

```suggestion
    found_part_from_config = any(
        part_from_config.lower() in feature.lower()
        for part_from_config in allowed_parts_from_config
        if allowed_parts_from_config
    )
```

src/extensions/score_metamodel/tests/rst/conf.py
```diff
@@ -20,7 +20,7 @@
     "sphinx_needs",
     "score_metamodel",
```
AlexanderLanin, line 21: _AI-assisted review (Claude):_

Adding `required_in_id = ["blabla"]` to the shared `conf.py` affects **every test that uses this Sphinx build**, not just the new `id_contains_feature` tests. Any test whose need ID happens to contain the string `"blabla"` would silently bypass the `id_contains_feature` check.

More importantly, this is a globally-permissive sentinel value chosen purely for testing. When reviewing or debugging failures, it's non-obvious why `blabla` is allowed globally. A more precise approach: add a dedicated `conf.py` under the `id_contains_feature/` subdirectory so test isolation is explicit.

docs/how-to/write_docs.rst
```diff
@@ -70,3 +70,47 @@ For further documentation on needextends please `look here <https://sphinx-needs
...
+     - ❌
+     - ``blabla`` has no relation to the path ``internals/safety/fmea``
+
+To explicitly allow a feature part that intentionally doesn't match the path
+(e.g. in a single module repository), add a matching string to ``required_in_id`` in ``conf.py``:
+
+.. code-block:: python
+
+   # conf.py
+   required_in_id = ["persistenc"]
```
AlexanderLanin, line 116: _AI-assisted review (Claude):_

`"persistenc"` is a partial string that would match `persistence`, `persistency`, etc. If this substring match is intentional, it should be explained in the docs — readers will assume it's a typo for `"persistence"` and either copy it incorrectly or fix it in their own `conf.py` without understanding why the truncation was used.

If the intent is an exact feature name, use the full string. If partial matching is a deliberate affordance, document it explicitly here.

src/extensions/score_metamodel/tests/rst/id_contains_feature/test_id_contains_feature.rst
```diff
@@ -34,3 +36,26 @@
...
+
+.. feat_req:: Testing if warning correctly triggers
+   :id: feat_req__id_contains__testing
+
+
+.. Testing if additional allowed param in conf.py is valid
+#EXPECT-NOT[+2]: Feature part
+
+.. feat_req:: Testing conf.py parameter
+   :id: feat_req__blabla__testing
```
AlexanderLanin, line 61: _AI-assisted review (Claude):_

This test verifies that `feat_req__blabla__testing` does not produce a warning when `required_in_id = ["blabla"]` is set — good. But there's no test for the **negative boundary**: a need whose ID contains `blabla` only in the type or title segment (e.g. `feat_req__other__blabla_testing`) should still trigger a warning.

Once the match is narrowed to the feature part (see comment on `id_contains_feature.py`), a test like this is needed to confirm the fix doesn't over-suppress.

AlexanderLanin (commented):

**Critical: 0 | Important: 2 | Suggestions: 1**

The feature direction is right — a config-driven escape hatch for single-module repos is a reasonable solution to #596. The implementation has two important correctness issues and one suggestion.

**Important: config match is against the full ID, not the feature part**
Line 63 checks `part_from_config.lower() in need.get("id")`, which matches the entire ID string (type prefix + feature + title). This means `required_in_id = ["feat"]` or `required_in_id = ["req"]` would silently disable the check for every requirement in the build. The match should be against `feature` (`parts[1]`) — the documented scope. See inline comment.

**Important: `required_in_id = ["blabla"]` added to the shared test `conf.py`**
This setting applies to all Sphinx test builds under `tests/rst/`, not just the `id_contains_feature` tests. Other test files inherit this config and would silently bypass the check for any ID containing `blabla`. Isolating this behind a per-subdirectory `conf.py` is cleaner and safer.

**Suggestion: docs example uses `"persistenc"` (truncated string)**
The example in `write_docs.rst` uses a partial string. Since the check does substring matching, this works — but readers will assume it is a typo. Either document that partial strings are intentional, or use the full module name to avoid confusion.

src/extensions/score_metamodel/tests/rst/conf.py
MaximilianSoerenPollak: This is a non issue as we only tests single checks in each RST file. 
This will not interfeer.

src/extensions/score_metamodel/tests/rst/id_contains_feature/test_id_contains_feature.rst
MaximilianSoerenPollak: Seems plausible. Will add this.

AlexanderLanin (approved):

PR #609, opened 2026-06-22
fix: needpie coverage

a-zw:
Coverage needpie diagram computed the pies incorrectly.


MaximilianSoerenPollak (approved):
Seems good to me. 💯

PR #614, opened 2026-06-24
test: Refactor RST tests

MaximilianSoerenPollak:
## 📌 Description
Completely refactor RST tests on how they work. 
Added test_metadata to ensure linking is possible in them as well

## 🚨 Impact Analysis

## ✅ Checklist


src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -227,107 +128,204 @@ def filter_warnings_by_position(
...
-
-    # Build the documentation with the enabled checks
-    app.config.score_metamodel_checks = rst_data.enabled_checks
+    ### Build the given rst file with Sphinx and check expected/unexpected warnings.
+    rst_data = count_need_objects(RST_DIR / rst_file)
+
+    # Build the documentation
+    app = sphinx_app_setup(RST_DIR / rst_file)
+    monkeypatch.chdir(app.srcdir)  # Sphinx resolves paths relative to the source dir
     app.build()
```
AlexanderLanin, line 279: _AI-assisted review (Claude):_

**Critical: check filter is set _after_ `app.build()` — checks run with the wrong filter**

The `score_metamodel_checks` config is set from the `test_metadata` need at line 318–319, but `app.build()` is called at line 308. The metamodel check runner (`_run_checks`) fires on the `build-finished` event, which is emitted *inside* `app.build()`. So the check filter is always empty when checks actually run — every file runs all checks, not just the one declared in `:check:`.

The old code did this correctly: `app.config.score_metamodel_checks = rst_data.enabled_checks` was set before `app.build()`.

The fix requires reading the `:check:` value before building. One approach: parse it from the RST source (the old `#CHECK:` comment pattern) or use a two-pass approach (build once to discover metadata, then build again with the filter). Alternatively, keep the `:check:` value in `conf.py` and pass it via `confoverrides` to `SphinxTestApp`.

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -227,107 +128,204 @@ def filter_warnings_by_position(
...
-                    pytest.fail(msg, pytrace=False)
+    # Get & parse metadata needs
+    needs_data = SphinxNeedsData(app.env)
+    needs_view = needs_data.get_needs_view()
+    extends = needs_data.get_or_create_extends().values()
+    all_needs = list(needs_view.values()) + list(extends)
+    _validate_need_count(all_needs, rst_data)
+
+    metadata_need = _get_test_metadata_need(needs_view, rst_data)
+    rst_data.enabled_checks = metadata_need.get("CHECK")
```
AlexanderLanin: _AI-assisted review (Claude):_

**Critical: field accessed as `"CHECK"` but registered as `"check"` — always returns `None`**

`NeedItem.get()` uses the registered field name as the key. The field is registered as `"check"` (lowercase) in `conf.py`'s `add_options_regex` and `add_needs_fields`, and `parse_test_metadata` correctly reads it as `need.get("check")` at line 163. But here the raw `NeedItem` is queried with uppercase `"CHECK"`, which will always return `None`.

The consequence compounds finding #1: even if the build order were fixed, the check filter would still be empty because this key lookup fails silently.

Fix: use `metadata_need.get("check")` (lowercase), or better, use the already-parsed `rst_data.metadata["check"]` which is set two lines later.

src/extensions/score_metamodel/tests/rst/id_contains_feature/test_id_contains_feature.rst
```diff
@@ -12,57 +12,65 @@
    # SPDX-License-Identifier: Apache-2.0
    # *******************************************************************************
 
-#CHECK: id_contains_feature
+
+.. test_metadata::
+   :id: test_metadata__id_contains_feature
+   :partially_verifies_list: tool_req__docs_common_attr_id_scheme
+   :test_type: requirements_test
```
AlexanderLanin: _AI-assisted review (Claude):_

**Important: `"requirements_test"` is not a valid `TestType` — runtime `KeyError` guaranteed**

The `TestType` literal in `attribute_plugin.py` accepts: `"fault-injection"`, `"interface-test"`, `"requirements-based"`, `"resource-usage"`. The value here is `requirements_test` (underscore, no hyphen, wrong suffix), which matches none of them.

`_build_test_properties` receives this as `test_type` and passes it directly to `pytest.mark.test_properties`. Depending on downstream validation, this either silently produces wrong XML output or fails at runtime when the report generator validates the value.

Fix: change to `requirements-based`.

src/extensions/score_metamodel/BUILD
```diff
@@ -82,6 +82,7 @@ score_pytest(
         "tests/rst/conf.py",
         "tests/rst/needs.json",
     ],
+    args = ["-s", "-vv"],
```
AlexanderLanin: _AI-assisted review (Claude):_

**Important: debug flags `-s -vv` left in all five `score_pytest` BUILD targets**

`-s` disables stdout capture (dumps all Sphinx output to terminal) and `-vv` enables double-verbose mode. These are debugging aids that should not be committed to CI targets — they produce noisy output and slow down iterative CI runs.

Remove `args = ["-s", "-vv"]` from all five `score_pytest` rules.

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -227,107 +128,204 @@ def filter_warnings_by_position(
...
+        "check": need.get("check"),
+    }
+    assert metadata["fully_verifies"] or metadata["partially_verifies"], (
+        f"Error in need: {need.get('id')}. "
+        "Either fully_verifies or partially_verifies must be provided"
+    )
+    return metadata
+
+
+def apply_test_metadata_props(
```
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion: `apply_test_metadata_props` duplicates `apply_test_metadata` from `attribute_plugin.py`**

`attribute_plugin.py` already exports `apply_test_metadata` (added in this PR) with the same purpose and the same `_build_test_properties` call. The local `apply_test_metadata_props` here does the same thing with slightly different file-path concatenation logic.

Consider using `apply_test_metadata` from the plugin directly (possibly extending its signature to accept a `file_prefix` for the path concatenation), rather than maintaining a second implementation that can drift.

src/extensions/score_metamodel/tests/rst/conf.py
```diff
@@ -15,20 +15,106 @@
...
+            "optional_links": {},
+        }
+    ]
+
+    # At parse time these are None, as they need to be filled post-parsing when all types are available to resolve the link targets.
+    # The dict maps the link name to a list of need types that are allowed as targets for that link.
+
+    changed_needs_types = list(test_metadata_need_type)
+    all_needs_types = app.config.needs_types
+    if "test_metadata" not in all_needs_types:
```
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion: `"test_metadata" not in all_needs_types` always evaluates the string against a list of dicts**

`all_needs_types` is a list of `NeedType` dicts (each with a `"directive"` key). The expression `"test_metadata" not in all_needs_types` checks whether the *string* `"test_metadata"` is an element of the list — it never matches a dict, so this guard is always `True` and the setup body always runs.

The intent (based on `__init__.py` line 224) is to check whether any existing type already has the directive:
```python
if "test_metadata" not in {nt["directive"] for nt in all_needs_types}:
```
Also: `app.config.needs_types = changed_needs_types` inside the loop overwrites the config on every iteration — it should be moved outside the loop.

AlexanderLanin (changes requested):

**Critical: 2 | Important: 1 | Suggestions: 2**

The refactor has solid motivation — moving test metadata from ad-hoc `#CHECK:` / `#EXPECT` comments into first-class `test_metadata` needs is a real improvement. The code structure is cleaner and the approach of attaching traceability links to RST-based tests is the right direction.

However there are two critical regressions that break the core mechanism:

1. **`app.build()` is called before `score_metamodel_checks` is set** (line 308 vs 318–319). The check filter fires inside `build-finished`, so every test file runs _all_ checks, not just the one declared in `:check:`. The old code set the config before building. This needs to be read from RST before the app is constructed — either by retaining a lightweight pre-parse of the `:check:` value, or by threading it through `confoverrides` in the `SphinxTestApp` constructor.

2. **`metadata_need.get("CHECK")` uses the wrong case** — the field is `"check"` (lowercase). This always returns `None`, compounding issue #1 (the filter is always empty).

Together these mean the check-scoping that was present in the old code is completely lost in this PR.

One additional important issue: the `requirements_test` value in `test_id_contains_feature.rst` is not in the `TestType` literal — the valid value is `requirements-based`.

The `-s -vv` debug args committed to all five BUILD targets should also be removed before merge.

AlexanderLanin (changes requested):

Re-review after new commit (faefabde). The only change since the last review is an addition to `docs/internals/extensions/rst_filebased_testing.md`.

**None of the previously reported issues have been addressed.** All five original findings remain open:

- ❌ **Critical** — `app.build()` called before `score_metamodel_checks` is set (line 308 vs 318)
- ❌ **Critical** — `metadata_need.get("CHECK")` uses wrong case; field is `"check"` (line 318)
- ❌ **Important** — `test_id_contains_feature.rst` uses `requirements_test`, not the valid `requirements-based`
- ❌ **Important** — `-s -vv` debug flags left in all five BUILD targets
- ❌ **Suggestion** — `apply_test_metadata_props` duplicates `apply_test_metadata` from `attribute_plugin.py`
- ❌ **Suggestion** — `conf.py` guard `"test_metadata" not in all_needs_types` checks a string against a list of dicts; always true

The two Criticals are blockers — with the current code the check filter is never applied during the build, so the per-file check scoping introduced by the `test_metadata` need has no actual effect.

src/extensions/score_metrics/sphinx_filters.py
```diff
@@ -207,6 +207,7 @@ def get_metrics_with_first_value_total(
     results.clear()
     # As kwargs ordering is deterministic this will always put the first total into results[0]
     _get_key_values(results, [str(value) for value in kwargs.values()])
+
```
a-zw, line 210: Why are there random changes like this one?

MaximilianSoerenPollak: Formatting. 
That was a change the pre-commit did when I ran `pre-commit run -a`.

src/extensions/score_metamodel/tests/test_rules_file_based.py
MaximilianSoerenPollak: Fixed.

src/extensions/score_metamodel/tests/rst/conf.py
MaximilianSoerenPollak: Fixed.

src/extensions/score_metamodel/tests/test_rules_file_based.py
MaximilianSoerenPollak: Instead enabled all checks as this is not possible.

src/extensions/score_metamodel/tests/rst/id_contains_feature/test_id_contains_feature.rst
```diff
@@ -12,57 +12,64 @@
...
```
AlexanderLanin: _AI-assisted review (Claude):_

**Important: `"requirements_test"` invalid `test_type` — still not fixed**

`test_id_contains_feature.rst` line 19 still has `:test_type: requirements_test`. The valid `TestType` values are `"fault-injection"`, `"interface-test"`, `"requirements-based"`, `"resource-usage"`. This value will be written to the JUnit XML as-is and silently produces a non-conforming test result.

Fix: change to `requirements-based`.

src/extensions/score_metamodel/tests/rst/conf.py
```diff
@@ -15,20 +15,106 @@
...
+            "description": "List of requirements that this RST test partially verifies",
+            "schema": {
+                "type": "array",
+                "items": {"type": "string"},
+            },
+            "nullable": True,
+        },
+    }
+    add_options_regex = {
+        "check": "^.*$",
```
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion: `:check:` field is dead code — registered but never used**

The `"check"` field is defined in `add_needs_fields` and `add_options_regex` in `conf.py`, and `parse_test_metadata` stores it (`need.get("check")`). But after removing `:check:` from all RST files (commit 0b56605), nothing reads `metadata["check"]` — neither `test_rules_file_based.py` nor `apply_test_metadata`. 

Remove it from `add_needs_fields`, `add_options_regex`, `parse_test_metadata`, and the `test_metadata` `optional_options` to avoid confusion. If per-file check scoping is re-introduced later, it should be re-added intentionally.

docs/internals/extensions/rst_filebased_testing.md
```diff
@@ -4,111 +4,159 @@
...
+> **Note:** RST comments used to separate or describe cases (for example
+> `.. All required options are present`) must not contain `::`, otherwise the
+> parser counts them as needs.
 
 ## Graph check tests
 
 Graph checks (defined in `metamodel.yaml` under `graph_checks:`) are all executed
-by a single Python function `check_metamodel_graph`. Use that function name in `#CHECK:`:
+by a single Python function `check_metamodel_graph`. Use that function name in the
+`test_metadata` need's `:check:` option:
```
AlexanderLanin: _AI-assisted review (Claude):_

**Suggestion: documentation still describes `:check:` as a working feature**

Lines 44–46 describe `<check functions>` as a field the user can set, and lines 106–114 show a `.. test_metadata::` example with `:check: check_metamodel_graph`. But `:check:` was removed from all RST test files in commit 0b56605, and the test runner no longer reads it. A contributor following this doc would add `:check:` to their RST file and wonder why check scoping has no effect.

Either remove the `:check:` references from the doc (and the example), or re-implement the feature properly before documenting it.

AlexanderLanin (changes requested):

Re-review after commits 0b56605–9ca8937.

**Critical: 0 | Important: 1 | Suggestions: 2**

Good progress — both critical build-ordering issues from the first review are resolved by removing the per-file `:check:` scoping entirely (commit 0b56605). The debug flags, duplicate function, and `conf.py` guard are also fixed. The PR is substantially better.

**Still open from previous reviews:**

- ❌ **Important** — `:test_type: requirements_test` in `test_id_contains_feature.rst` is still there. `requirements_test` is not a valid `TestType` literal. Change to `requirements-based`.

**New findings:**

- ⚠️ **Suggestion** — The `"check"` field is now dead code: registered in `conf.py`, populated in `parse_test_metadata`, but never read by anything since `:check:` was removed from all RST files. Clean it up.
- ⚠️ **Suggestion** — The documentation still describes `:check:` as a feature (lines 44–46 and 106–114 of `rst_filebased_testing.md`), with example code blocks showing how to use it. Since the feature no longer works, this will mislead contributors. Either remove the references or re-implement the feature.

Other issues (typo in `descdription`, `test_need_extends.rst` missing `test_metadata`, `safety: YES` invalid values) are already flagged by other reviewers — I won't duplicate them.

AlexanderLanin (commented):

Re-review after commit 9628353f.

**Critical: 0 | Important: 1 | Suggestions: 1**

Good progress — five findings from the previous review are now resolved:

- ✅ `requirements_test` → `requirements_based` fixed in `test_id_contains_feature.rst`
- ✅ `:check:` dead-code field removed from `conf.py`
- ✅ Doc section describing `:check:` as a working feature removed
- ✅ `descdription` typo fixed in `_get_default_metadata_need()`
- ✅ `.rst` extension now appended to `file_name` in `test_rules_file_based.py`

Two findings remain open:

**Important — `test_need_extends.rst` has no `test_metadata` block and uses invalid `safety` values**

This file still has no `.. test_metadata::` directive (existing thread on line 16), so junitxml `file`/`line` attributes fall back to defaults for this test. It also still sets `:safety: YES`, `:safety: NO`, and `:+safety: YES` (lines 28, 34, 56, 72) on `stkh_req` needs — these violate the `QM|ASIL_B` constraint and will generate unrelated format warnings from `check_extra_options`, adding noise to the warning list the test asserts against. Both issues are flagged in existing threads and remain unaddressed.

**Suggestion — `:expect_not: title contains` is a vacuous assertion (`test_prohibited_words.rst` lines 40, 55)**

The actual warning from `check_for_prohibited_words` reads `contains a weak word`, not `title contains`. The substring `title contains` never appears in any warning, so these `:expect_not:` assertions always pass and provide no safety net. Fix: change to `:expect_not: contains a weak word` (flagged in existing threads on lines 41 and 56).

AlexanderLanin (approved):

PR #623, opened 2026-06-26
docs: more requirements and links

a-zw:
## 📌 Description

Some process requirements are already satisfied / implemented.


AlexanderLanin (approved):
LGTM. Detailed review as usual when we add versioned links.

Issue #594, opened 2026-06-10
Hard-deprecate :satisfies: for feat_req/comp_req after migration to :derived_from


RSingh1511:
satisfies is kept as a backward-compat optional link in feat_req and comp_req while downstream repos (score, [module_template](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) migrate to derived_from (introduced in PR #446, aligned with process_description#621). Once migration is complete, remove satisfies from feat_req.optional_links, comp_req.optional_links, graph_checks, and needs_extra_links in [metamodel.yaml](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

MaximilianSoerenPollak:
@RSingh1511 will you make a PR for this?

RSingh1511:
> [@RSingh1511](https://github.com/RSingh1511) will you make a PR for this?

Created- https://github.com/eclipse-score/docs-as-code/pull/616

AlexanderLanin:
implemented

PR #624, opened 2026-06-29
docs(requirements): improve traceability

a-zw:


src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -660,8 +664,10 @@ needs_types:
       # req-Id: tool_req__docs_common_attr_status
       status: ^(valid|invalid)$
     mandatory_links:
+      # req-Id: tool_req__arch_linkage_safety
       included_by: real_arc_int
     optional_links:
```
AlexanderLanin, line 669: _AI-assisted review (Claude):_

**Important — graph_check silently narrows safety enforcement**

Removing `satisfies: safety == QM` from `tool_req__docs_common_attr_safety_link_check` is safe *if and only if* no `feat_req`/`comp_req` in any downstream repo still uses `:satisfies:` instead of `:derived_from:`. But issue #594 is still **open**, meaning the downstream migration may not be complete.

If a downstream repo has:
```rst
.. feat_req:: ...
   :safety: QM
   :satisfies: some_asil_gd_req
```
that repo will now pass the graph check silently — it used to be rejected by the `satisfies: safety == QM` clause.

Before this merges, either:
1. Confirm all downstream repos (`score`, `module_template`) have migrated to `derived_from`, then close #594; or
2. Keep `satisfies: safety == QM` in the graph check until migration is confirmed complete.

The test added in `test_options_extra_option.rst` only verifies that `satisfies` is rejected as an *extra option* for `feat_req` in this repo's own metamodel tests — it doesn't cover the graph_check enforcement for downstream repos that still emit `:satisfies:`.

AlexanderLanin (commented):

**Critical: 0 | Important: 2 | Suggestions: 1**

Overall: two logically separate changes bundled in one PR — the `satisfies` hard-deprecation for `feat_req`/`comp_req` (commit 46cfe04) and the new `tool_req__arch_linkage_safety` requirement. The metamodel changes are clean and the traceability annotations look correct. The main concerns are a silent safety-check gap introduced by the graph_check change and a consistency issue in the requirement table itself.

The existing Copilot inline comments (on `consist_of` vs `consists_of`, `included by` vs `included_by`, grammar) are correct and should be addressed before merge.

docs/internals/requirements/requirements.rst
```diff
@@ -788,6 +791,34 @@ Architecture Attributes
...
+  :id: tool_req__arch_linkage_safety
+  :implemented: PARTIAL
+  :version: 1
+  :satisfies: gd_req__arch_linkage_safety[version==1]
+  :parent_covered: YES
+
+  .. csv-table::
+     :header: "Link source", "Relation", "Link Target", "Mandatory", "Implemented"
+
+     feat, consist_of, comp, yes, no
```
AlexanderLanin: _AI-assisted review (Claude):_

**Important — requirement table row 1 documents a non-existent link**

The row `feat, consist_of, comp, yes, no` says this link is mandatory but not yet implemented. However `feat` has no `consists_of` (or `consist_of`) in its `optional_links` or `mandatory_links` in `metamodel.yaml` — the link simply doesn't exist yet, not just unimplemented.

The warning below acknowledges the `consist_of` vs `consists_of` spelling discrepancy against `gd_req__arch_linkage_safety`, but the more fundamental issue is that this row documents a planned-but-absent link with no corresponding metamodel entry or `req-Id` annotation on `feat`.

Consider either:
1. Adding a `# TODO` comment under `feat.optional_links` annotated with `req-Id: tool_req__arch_linkage_safety` to make the gap traceable in the metamodel (even if the link target type is TBD), or  
2. Noting explicitly in the table that this row is aspirational (documenting `gd_req__arch_linkage_safety`'s intent, not the current implementation).

As written, a reader auditing `feat`'s metamodel definition will find no evidence of this mandatory link.

docs/internals/requirements/requirements.rst
```diff
@@ -788,6 +791,34 @@ Architecture Attributes
...
      "includes", "corresponding architecture element lower level"
 
+The following requirement may be overlapping with other tool requirements,
+but for ease of traceability this is a separate one.
+
+.. tool_req:: Correlations of the architectural building blocks
+  :id: tool_req__arch_linkage_safety
+  :implemented: PARTIAL
+  :version: 1
+  :satisfies: gd_req__arch_linkage_safety[version==1]
```
AlexanderLanin, line 801: _AI-assisted review (Claude):_

**Suggestion — add `version` to `:satisfies:` entries for `valid_from`/`valid_until`**

The existing requirement `tool_req__docs_req_attr_validity_correctness` was already satisfying `gd_req__req_validity[version==1]`. The two new entries added here:

```
gd_req__req_attr_valid_from[version==1],
gd_req__req_attr_valid_until[version==1],
```

look correct. Just confirming the `[version==1]` pin is intentional here (locking to the current version of those GD reqs). If those GD reqs are expected to evolve this may need updating, but that's not unusual for this codebase.

docs/internals/requirements/requirements.rst
a-zw: >  the link simply doesn't exist yet, not just unimplemented.

That's the same thing. @AlexanderLanin educate your clanker a bit?

AlexanderLanin (commented):

The fixup commits (especially `e4ac0aa7` "fix: consistently consists_of") have resolved all previously flagged issues: the `consist_of`/`consists_of` inconsistency is fixed, the `included by`/`included_by` inconsistency is fixed, the warning text with grammar issues was removed, and the `"maybe be"` typo is gone. The `[version==1]` constraints are present on the new `:satisfies:` entries.

The requirement table is correct against the current metamodel:
- `feat, consists_of, comp, yes, no` — accurately documents a planned-but-unimplemented mandatory link; the `req-Id: tool_req__arch_linkage_safety` annotation already covers the `feat.optional_links` block in the metamodel
- `feat, includes, logic_arc_int, yes, only optional` — accurately documents the mismatch between the process requirement and the current metamodel implementation

One minor observation: rows `comp, implements, logic_arc_int` and `comp, uses, logic_arc_int` only list `logic_arc_int` as the link target, but the metamodel also allows `real_arc_int_op` for both. This may be intentional (the table documents the safety-relevant subset from `gd_req__arch_linkage_safety`), but if completeness matters, adding `real_arc_int_op` rows would make the table a full mirror of the metamodel.

The existing open inline comments from Copilot/other reviewers are stale — all refer to issues resolved by the subsequent fixup commits. This PR looks ready to merge.

AlexanderLanin (approved):

PR #631, opened 2026-07-06
Require valid_from in validity consistency check

antonkri:
Require and normalize valid_from in validity consistency check

### DIFF
--- a/src/extensions/score_metamodel/metamodel.yaml
+++ b/src/extensions/score_metamodel/metamodel.yaml
@@ -263,6 +263,7 @@ needs_types:
       rationale: ^.+$
       # req-Id: tool_req__docs_common_attr_security
       security: ^(YES|NO)$
+      valid_from: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)(\.(0|[1-9]\d*))?$
     optional_options:
       # req-Id: tool_req__docs_req_attr_reqcov
       reqcovered: ^(YES|NO)$
@@ -270,7 +271,6 @@ needs_types:
       testcovered: ^(YES|NO)$
       hash: ^.*$
       # req-Id: tool_req__docs_req_attr_validity_correctness
-      valid_from: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)(\.(0|[1-9]\d*))?$
       valid_until: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)(\.(0|[1-9]\d*))?$
     tags:
       - requirement
@@ -292,6 +292,7 @@ needs_types:
       status: ^(valid|invalid)$
       # req-Id: tool_req__docs_common_attr_description
       content: ^[\s\S]+$
+      valid_from: ^v(0|[1-9]\d*)\.(0|[1-9]\d*)(\.(0|[1-9]\d*))?$
     optional_links:
       # req-Id: tool_req__docs_req_link_satisfies_allowed
       derived_from: stkh_req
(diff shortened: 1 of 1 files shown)


src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -278,6 +278,16 @@ def parse_milestone(value: str) -> tuple[int, int, int]:
     return (major, minor, patch)
 
 
+def _normalize_validity_attr(value: object | None) -> str | None:
```
MaximilianSoerenPollak: I would prefer if this would be a generic function (name) instead of one specialized for validity attributes. 

For the code  itself, is there a scenario where the value is NOT an int/float or a string? 
Like could it be a list or so? 
If that is the case we would return just the list as a string here, not sure that is what we want?
So we probably should program more defensively to error in such scenarios too.

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -286,17 +296,24 @@ def check_validity_consistency(
...
+    valid_until = _normalize_validity_attr(need.get("valid_until", None))
+
+    # Temporary grace period: only enforce presence of valid_from as non-fatal info.
+    if not valid_from:
+        msg = "missing validity attribute: valid_from must be set."
+        log.warning_for_need(need, msg, is_new_check=True)
+        return
 
-    if not valid_from or not valid_until:
+    if not valid_until:
```
MaximilianSoerenPollak: There is no error msg here, but there is one in valid_from? Is this an oversight or on purpose?

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -291,12 +291,19 @@ def check_validity_consistency(
...
         return
 
-    valid_from = need.get("valid_from", None)
-    valid_until = need.get("valid_until", None)
+    valid_from_values = _get_normalized(need, "valid_from")
+    valid_until_values = _get_normalized(need, "valid_until")
 
-    if not valid_from or not valid_until:
+    if valid_until_values and not valid_from_values:
+        log.warning_for_need(need, "is missing required attribute: `valid_from`.")
```
MaximilianSoerenPollak: A bit confused here. 
According to metamodel.yaml valid_from and valid_until are both optional options. 
Here now one is protrayed as mandatory option. 
Did you forget to add the change to the metamodel.yaml?

antonkri: Hey Maximilian, no. What I want to achieve is that everyone do the scoping of their requirements and tell us, which requirements are valid for release score v1.0 and which for later releases. In my opinion, valid_until is not mandatory, as the requirement can be valid forever. But valid_from is more important as we need to know, when the implementation is available for the first time and can be e.g. tested.

antonkri: ...sorry The metamodel was changes in the previous commit, please check

src/extensions/score_metamodel/checks/check_options.py
FScholPer, line 1: Please set the flags to mandatory in the metamodel and add the checks as implemented in tool requirements

FScholPer (changes requested):

FScholPer (dismissed):

src/extensions/score_metamodel/checks/check_options.py
```diff
@@ -291,12 +291,19 @@ def check_validity_consistency(
...
         return
 
+    if not valid_from_values or not valid_until_values:
```
MaximilianSoerenPollak: This is still an issue I think, as valid_from_values is mandatory but `valid_until_values` is not. 
Therefore if valid_until_values is not set this will still enter and exit early

MaximilianSoerenPollak (approved):

PR #633, opened 2026-07-08
feat: build host-only doc tools in host config under cross --platforms

umaucher:
## Why

The runnable documentation targets created by `docs()` — `docs`, `docs_combo`, `docs_check`, `docs_link_check`, `live_preview`, `live_preview_combo_experimental`, `ide_support` — are host tools that run Python/Sphinx on the build machine. When a consumer builds with a cross `--platforms` (e.g. `--config=per-x86_64-qnx -- //...`), these top-level targets get configured for the **target** platform and fail Python toolchain resolution (`No matching toolchains ... tools/python`), breaking the whole build in cross-compiling consumers (persistency, reference_integration).

## What

Each host-only run target is wrapped in a small rule (`_host_binary`) whose **incoming transition** pins `//command_line_option:platforms` to whatever `--host_platform` is set to. The targets therefore always build and run in the host configuration regardless of the top-level `--platforms`.

- The wrapper forwards the executable (via symlink), runfiles, and `RunEnvironmentInfo`, so `bazel run //:docs` (and the `env`-carrying targets) keep working unchanged.
- The transition follows `--host_platform` dynamically rather than hardcoding a platform, so it works for any consumer's host.
- `sphinx_build` is only ever consumed as a tool (already built in the exec/host config), so it needs no transition; it is tagged `manual` so a wildcard `bazel build //...` under a cross platform skips it instead of trying to configure it for the target platform.

## Validation

- Host: `bazel build //:docs //:needs_json //:metrics_json`, `bazel run //:ide_support` — no regression.
- Cross: in a consumer with a local override, `bazel build --config=per-x86_64-qnx -- //...` no longer fails Python toolchain resolution — the wrapped run targets transition to the host config, while `sphinx_build` (a `manual` tool) is skipped by the wildcard and built only in the exec/host config when consumed by `needs_json`. `cquery --transitions=lite` confirms the wrapped targets transition to the host config. (The only remaining consumer failure is the unrelated `valid_from` metamodel skew fixed by eclipse-score/score#3064.)
- `buildifier` clean.

No tracking issue.

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -45,6 +45,71 @@ load("@aspect_rules_py//py:defs.bzl", "py_binary", "py_venv")
 load("@docs_as_code_hub_env//:requirements.bzl", "all_requirements")
 load("@rules_python//sphinxdocs:sphinx.bzl", "sphinx_build_binary", "sphinx_docs")
 
+# -- Host-platform transition for host-only documentation tooling --------------
+#
+# The runnable doc targets (docs, ide_support, live_preview, ...) are pure host tools that run
+# Python/Sphinx locally. Under a cross `--platforms` (e.g. QNX) they would be configured for the
+# target platform and fail Python toolchain resolution. Instead of `target_compatible_with` (which
+# would only make `bazel build //...` skip them), each is wrapped in a rule whose incoming transition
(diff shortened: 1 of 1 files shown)


umaucher:
requires
- eclipse-score/score#3064

PR #634, opened 2026-07-08
Make version mandatory & further fixes regarding process

MaximilianSoerenPollak:
This is making the version mandatory. 

The other changes in the tool_requirements etc. are needed to enable building for the newest process commit done as a git_override here. 

The git_override should be removed once release is done from process


a-zw (approved):
Two consumer tests failing. That is expected because this is a breaking change. More surprising that "only two" fail.

dcalavrezo-qorix (approved):

PR #644, opened 2026-07-16
#570 FIX broken links

RamakrishnanPK:
[TASK] Add broken link fix extension to ignore errors

 - Correct the broken links to the latest links available.


RamakrishnanPK:
Hi @MaximilianSoerenPollak,

Request to review and provide your comments.

Currently the checks were failing due issues on other repos via #https://github.com/eclipse-score/docs-as-code/pull/634 
Create a ticket to fix the issue with process_description repo: https://github.com/eclipse-score/process_description/issues/749

MaximilianSoerenPollak:
I'm not sure if I like the fix being a sphinx extension but I honestly do not have a better idea here.

RamakrishnanPK:
> I'm not sure if I like the fix being a sphinx extension but I honestly do not have a better idea here.

## 1. 403 client errors - due to server not allowing bot requests

### user_agent
`
user_agent = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36'
`
User agent configuration did not work!

### linkcheck_ignore

```
linkcheck_ignore = [
    r'https://example\.com/broken-page',  # Ignores a specific URL
]
```

but this will ignore any real broken links as well. ( I mean 404 errors). I have to only ignore check only for 403 errors.

## 2. .md file referenced in the documentation with anchors. 

Will have to wait until those are published / released as .html pages. but this is not in our control. 

So, I do not have an option either! Urgency is BOT would again create a "Broken Links" automated issue!

BUILD
```diff
@@ -38,3 +38,9 @@ alias(
     name = "actionlint",
     actual = "@score_devcontainer//tools:actionlint",
 )
+
+# bazel run //:lint_check
+alias(
+    name = "link_check",
+    actual = ":docs_link_check",
+)
```
AlexanderLanin: isnt this alias just confusing? None of the other docs targets have an alias.

AlexanderLanin:
does this replace .github/workflows/link_check.yml?

BUILD
RamakrishnanPK: I have added this before start of the activity. which now is outdated. 

have removed this in the next commit.  #612fc22

RamakrishnanPK:
image

build output - reference.

MaximilianSoerenPollak (approved):

PR #645, opened 2026-07-16
Fix: remove root location enforcement

MaximilianSoerenPollak:
## 📌 Description
We currently enforce the docs macro to only be used in the root build file for no reason as far as I can see. 
Removing this will open up a usecase of having multiple docs marcos in the same project.

## 🚨 Impact Analysis

## ✅ Checklist


MaximilianSoerenPollak:
This most likely actually will not fix the issue I think. 
We should instead add / allow for the `name` attribute as is common in bazel macros.

PR #648, opened 2026-07-17
Msp allow docs name

MaximilianSoerenPollak:
## 📌 Description
Allow the docs macro to be used with a `name` attribute as is custom for almost all bazel macros. 
This will allow us in the future to use the docs marco in the same build file multiple times, with different inputs and have no collision. 
This should be a good first step towards coming closer to how bazel macros should be written and make it more useable in other settings too. 

=> Still a challenge though is how we can enforce non idiotic usage of this in S-CORE. 

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -46,24 +46,36 @@ load("@docs_as_code_hub_env//:requirements.bzl", "all_requirements")
 load("@rules_python//sphinxdocs:sphinx.bzl", "sphinx_build_binary", "sphinx_docs")
 
 def _rewrite_needs_json_to_docs_sources(labels):
-    """Replace '@repo//:needs_json' -> '@repo//:docs_sources' for every item."""
+    """Replace '@repo//:needs_json' -> '@repo//:docs_sources' for every item.
+
+    Also handles a producer using a custom docs() `name`
+    (e.g. '@repo//:foo_needs_json' -> '@repo//:foo_docs_sources').
+    """
     out = []
     for x in labels:
         s = str(x)
         if s.endswith("//:needs_json"):
             out.append(s.replace("//:needs_json", "//:docs_sources"))
+        elif "//:" in s and s.endswith("_needs_json"):
+            out.append(s[:-len("needs_json")] + "docs_sources")
         else:
             out.append(s)
     return out
 
 def _rewrite_needs_json_to_sourcelinks(labels):
-    """Replace '@repo//:needs_json' -> '@repo//:sourcelinks_json' for every item."""
+    """Replace '@repo//:needs_json' -> '@repo//:sourcelinks_json' for every item.
+
+    Also handles a producer using a custom docs() `name`
+    (e.g. '@repo//:foo_needs_json' -> '@repo//:foo_sourcelinks_json').
+    """
     out = []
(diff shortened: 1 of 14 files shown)


src/extensions/score_metamodel/external_needs.py
```diff
@@ -48,7 +48,12 @@ def _parse_bazel_external_need(s: str) -> ExternalNeedsSource | None:
     repo, path_to_target = repo_and_path.split("//", 1)
     repo = repo.lstrip("@")
 
-    if path_to_target == "" and target in ("needs_json", "docs_sources"):
+    # The producer's docs() macro may have used a custom `name`, in which case its targets
```
MaximilianSoerenPollak, line 51: Probably can make this nicer with saying `"docs_sources" in target`

src/helper_lib/__init__.py
```diff
@@ -34,6 +35,22 @@ def config_setdefault(config: Config, name: str, value: Any) -> None:
...
+
+    Several of our extensions may be used either standalone or bundled together via
+    score_sphinx_bundle, in either load order, and may each want the same config value
+    registered. Sphinx raises if `add_config_value` is called twice for the same name,
+    so this guards the registration to make it safe to call from multiple extensions.
+    """
+    # Sphinx has no public API for this check either; `_options` is the internal dict
+    # `Config.add` itself consults to reject duplicate registrations.
+    if name not in app.config._options:  # pyright: ignore [reportPrivateUsage]
+        app.add_config_value(name, default, rebuild=rebuild)
```
MaximilianSoerenPollak: is this truly needed? 

Not using score_bundle should be not allowed in S-CORE, so I thin kwe can remove this,

MaximilianSoerenPollak (commented):
Probably 1 - 2 things still to iron out here, but overall looks like a good approach I think

docs.bzl
```diff
@@ -127,12 +139,21 @@ def _missing_requirements(deps):
         fail(msg)
     fail("This case should be unreachable?!")
 
-def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good = None, metamodel = None):
+def docs(name = "docs", source_dir = "docs", data = [], deps = [], scan_code = [], known_good = None, metamodel = None):
     """Creates all targets related to documentation.
 
     By using this function, you'll get any and all updates for documentation targets in one place.
 
+    The macro can be called multiple times in the same BUILD file as long as each call uses a
```
a-zw, line 147: I'm not convinced that this is a good idea.

There can only be exactly one documentation hosted by github.io. What is the motivation to have more?

The need to distribute documentation across the repo is addressed by implemented [DR-008-infra](https://eclipse-score.github.io/score/main/design_decisions/DR-008-infra.html).

MaximilianSoerenPollak: Yes there can be only one hosted, that is correct. 

This is to allow for multiple docs to be build, without it being hosted. 
Basically `many independent documentations in one repo` and allow / enable them all to be build without them needing to be hosted. 
This is also in anticipation of the `split up` we might need to do in regards to trlc things.

PR #650, opened 2026-07-20
chore: update to latest rules_python (2.2.0)

a-zw:


MODULE.bazel
```diff
@@ -19,7 +19,8 @@ module(name = "score_docs_as_code")
 # Python version
 #
 ###############################################################################
-bazel_dep(name = "rules_python", version = "1.8.3")
+bazel_dep(name = "rules_python", version = "2.2.0")
```
MaximilianSoerenPollak, line 22: Still needed then, directly? 
Or do we need it cause of `aspect_rules_py` ?

MaximilianSoerenPollak (approved):

MaximilianSoerenPollak:
Very suprised that none of the other repos is broken by this.l 
But I guess `root repo` takes precidence so they still use whatever `their` version is.

a-zw:
Only score_platform issues a warning:

>  UserWarning: score `bazel run //:docs_check`: WARNING: For repository 'rules_python', the root module requires module version rules_python@1.8.3, but got rules_python@2.2.0 in the resolved dependency graph. Please update the version in your MODULE.bazel or set --check_direct_dependencies=off

That repo already has plenty of warnings though (versions attribute mandatory).

MODULE.bazel
a-zw: We still need it for the Python toolchain setup and pip package management.

PR #657, opened 2026-07-22
feat: allow non root docs()

AlexanderLanin:
## 📌 Description

So far docs() was only callable from the root BUILD file. While that was intentional, this also prevents us from running docs() commands in order to test them properly.

Quite a lot of technical debt has accumulated due to this restriction we imposed. Notably:
* Finding needs_json from different places was not implemented
* Testlogs read everything from the module, but a non root docs() command will typically now know most of the linked requirements
* Generated documentation must not be placed into <root>/_bazel but in <docs()-dir>/bazel

## 🚨 Impact Analysis

- [x] This change does not violate any tool requirements
- [ ] ... is covered by existing tool requirements

## ✅ Checklist

- [ ] Added/updated documentation for new or changed features <-- intentionally skipped for now. Not sure we want that.


docs.bzl
```diff
@@ -137,16 +137,13 @@ def docs(source_dir = "docs", data = [], deps = [], scan_code = [], known_good =
       data: Additional data files to include in the documentation build.
       deps: Additional dependencies for the documentation build.
       scan_code: List of code targets to scan for source code links.
+      test_sources: Optional list of repo-relative directory paths which will be used to filter testcases for documentation generation.
```
MaximilianSoerenPollak, line 140: What happens to testcases that do not have `file` in the attributes? 
As that is added by our plugin I think?

src/extensions/score_metamodel/external_needs.py
```diff
@@ -31,10 +31,16 @@ class ExternalNeedsSource:
...
+    # True for a same-repo mount (`//pkg:needs_json`), whose runfiles live under
+    # `_main/…`. False for a cross-module mount (`@repo//…:needs_json`), whose
+    # runfiles live under `{bazel_module}+/…`.
+    is_local: bool = False
 
 
 def _parse_bazel_external_need(s: str) -> ExternalNeedsSource | None:
-    if not s.startswith("@"):
+    is_cross_module = s.startswith("@")
+    is_local = s.startswith("//")
```
MaximilianSoerenPollak, line 42: Is this leakage from the `sphinx-mounts` tryouts you had? 
This seems a bit unrelated at least from the comment I don't quiet get the relation?

src/extensions/score_source_code_linker/xml_parser.py
```diff
@@ -54,6 +55,38 @@
...
+    An empty `allowed_dirs` disables filtering (everything is in scope). Otherwise
+    a testcase is in scope only if it has a `file` attribute that is located under
+    one of the allowed (repo-relative) directories.
+    """
+    if not allowed_dirs:
+        return True
+    if not test_file:
+        return False
+    file_path = Path(test_file)
+    return any(file_path.is_relative_to(allowed) for allowed in allowed_dirs)
```
MaximilianSoerenPollak, line 87: If this is now new behaviour, and different than what currently is (have everything regardlesss of attributes & properties be a testcase due to the metrics) we should think if this is okay and if so it should be documented

src/extensions/score_source_code_linker/xml_parser.py
```diff
@@ -163,7 +196,9 @@ def parse_properties(case_properties: dict[str, Any], properties: Element):
     return case_properties
 
 
-def read_test_xml_file(file: Path) -> tuple[list[DataOfTestCase], list[str], list[str]]:
+def read_test_xml_file(
+    file: Path, allowed_dirs: list[str] | None = None
+) -> tuple[list[DataOfTestCase], list[str], list[str]]:
```
MaximilianSoerenPollak, line 201: For future cleanups someday. 

I truly don't like the return type here, seems so messy. Maybe a dataclass would be nicer. 
But that is sometime in the future, not an issue right now.

MaximilianSoerenPollak (commented):
Overall looks good, just some questions I have that we should address.

docs.bzl
AlexanderLanin: Yeah those would show up in all docs as unlinked tests. Acceptable for now / for my test-use-case.

src/extensions/score_metamodel/external_needs.py
AlexanderLanin: There are two changes:
* Since docs() does no longer need to be in root, we need to find the needs.json file in non root locations

And once we allow multiple docs() in one repository, they can depend on eachother.
* So the import of needs.json needs to handle local pathes as well, not only external pathes that start with `@repo`.

src/extensions/score_source_code_linker/xml_parser.py
AlexanderLanin: Default is unfiltered. The entire new docs()-parameter / behavior is undocumented.

Maybe we should call it experimental?

MaximilianSoerenPollak: Hmm I think we should do that, but in seperate PR's. 
We need to document the `docs` macro somewhere in the future.

MaximilianSoerenPollak (approved):
Looks like a good step in the right direction.
Talked about in calls & explained. So all fine for me atm.

PR #659, opened 2026-07-23
fix: Update tool requirements & update to process 2.0.2

MaximilianSoerenPollak:
## 📌 Description
Updates to newest process release, and adapts one tool requirement and expands it so it covers the new process req version

## 🚨 Impact Analysis

## ✅ Checklist


docs/internals/requirements/requirements.rst
```diff
@@ -273,20 +273,48 @@ Versioning
...
-   :parent_covered: NO: parent talks about setting covered to false, but we want to issue a build error.
-   :satisfies: gd_req__req_suspicious[version==2]
+   :version: 2
+   :parent_covered: YES: Together with :need:`tool_req__docs_req_testlink_outdated_check` and :need:`tool_req__docs_req_attr_testcov_update`
+   :satisfies: gd_req__req_suspicious[version==3]
    :status: invalid
 
-   Docs-as-Code shall check if linked parent needs have different versions, compared to
+   Docs-as-Code shall check if linked needs have different versions, compared to
    the version the need was originally linked to.
```
a-zw: Parent [gd_req__req_suspicious](https://eclipse-score.github.io/process_description//main/process_areas/requirements_engineering/guidance/requirements_process_reqs.html#gd_req__req_suspicious) says:

> Based on the requirement versioning it shall be checked if a requirement was updated but not the linked tests. In case an update was detected the attribute complete test coverage shall be set to “No”

I don't see how we cover this at all? This tool_req__docs_common_attr_suspicious is not about tests and it also sounds more general going beyond just tests.

docs/internals/requirements/requirements.rst
```diff
@@ -273,20 +273,48 @@ Versioning
...
+   :implemented: NO
+   :version: 1
+   :parent_covered: YES: Together with tool_req__docs_req_testlink_outdated_check tool_req__docs_common_attr_suspicious
+   :satisfies: gd_req__req_suspicious[version==3]
+   :status: invalid
+
+   Docs-as-Code shall set the ``testcovered`` attribute (see
+   :need:`tool_req__docs_req_attr_testcov`) of a requirement to ``No`` whenever
+   :need:`tool_req__docs_req_testlink_outdated_check` detects that a linked test is
+   outdated with respect to the requirement's version.
```
a-zw, line 304: I don't like this requirement. This would be very confusing for users, if the testcovered attribute in HTML is different than in the source rst.

a-zw (changes requested):

docs/internals/requirements/requirements.rst
MaximilianSoerenPollak: I see what you mean ye. 
This would be quiet confusing, we should find a way that we can force the user to update it.

docs/internals/requirements/requirements.rst
MaximilianSoerenPollak: I need to re-work this. 

What do we do with this one then, I think it makes sense to have, but maybe we don't need to write it down?

AlexanderLanin (dismissed):

Issue #674, opened 2026-07-29
Report cross-version metamodel compatibility findings in integrated documentation


AlexanderLanin:
## TL;DR / Executive summary

Cross-module documentation integration combines independently released modules, while Bazel resolves one current validation baseline for the complete graph. A module can therefore be locally valid at release time yet trigger new metamodel findings during integration.

Introduce a compatibility-aware validation mode for cross-module integration only:

- Keep metamodel warnings from top-level documentation fatal.
- Report, but do not fail on, metamodel warnings originating in imported `docs()` targets that have already passed their standalone validation.
- Keep concrete version-mismatch warnings fatal when both endpoints are local; report, but do not fail on, them when at least one endpoint is external.

Non-fatal compatibility findings must remain prominent in console output and on the generated top-level documentation index. Standalone validation remains strict and unchanged.

## Problem

S-CORE documentation is integrated from modules that evolve independently and are not upgraded in lockstep. As a result, different modules may support different versions of docs-as-code, the metamodel, and documentation capabilities.

During integration, Bazel resolves a single effective dependency graph. It selects the newest metamodel requested by any participating module and the newest set of feature requirements referenced by any participating module; that resolved set is applied to **all** features in the integrated build. The consuming documentation is therefore typically validated against a newer baseline than the one available when a supplying module was released. Consequently, a module that was valid when published may fail integration simply because the consumer applies newer validation rules. Although alignment processes exist, they are not yet applied consistently; and even with consistent use, their deliberately infrequent cadence—typically every two to eight weeks—means that temporary version skew remains expected.

Examples include attributes or links that became mandatory only in later metamodel versions, or changed version requirements. These are **compatibility findings**: validation findings that are expected and explainable from the producer/consumer version skew. They are classified separately so integration can continue while the compatibility gap remains clearly visible; they are not necessarily defects in the supplying module.

The existing behaviour effectively assumes a continuously integrated project landscape in which participating modules advance together. Definitions, processes, and alignment initiatives exist to synchronise independently released modules with different upgrade schedules, but they cannot be applied perfectly or eliminate version skew completely.

## High-level solution

Introduce a **compatibility-aware validation mode** for **cross-module documentation integration only**.

[...]

PR #676, opened 2026-07-31
feat: add opt-in cross-module compatibility warnings

AlexanderLanin:
## Summary

Today, any validation mismatch in an integrated documentation build aborts the build. This is correct and safe, but it prevents integration entirely until every participating module is perfectly aligned.

This feature allows integration to proceed during temporary version skew, but only for explicitly enabled, well-defined cross-module compatibility cases. It does not suppress arbitrary validation failures: local findings and every non-enabled category remain fatal.

Enabled compatibility findings stay highly visible: they are emitted in the build output, shown prominently on the generated documentation start page, and written to JSON and HTML reports for CI follow-up.

## Supported opt-in cases

- missing mandatory attributes in mounted or imported documentation modules
- missing mandatory links in mounted or imported documentation modules
- version mismatches in need links when at least one endpoint is external

Refs #674


docs/how-to/cross_module_compatibility.rst
```diff
@@ -0,0 +1,51 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2026 Contributors to the Eclipse Foundation
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+Cross-Module Compatibility Findings
```
a-zw: This is not a how-to. This belongs under docs/concept/.

a-zw (changes requested):

nradakovic (approved):

PR #684, opened 2026-08-03
feat: optional conf.py

AlexanderLanin:
Motivation: 90-100% of users don't know how changes to conf.py affect documentation build. They assume they can just add code there. While this may be helpful for some quick fixes or workarounds, we need to discourage people from touching the file. As long as the file is there, people will touch it. So let's get rid of the file!


docs.bzl
```diff
@@ -175,7 +204,36 @@ def docs(
     """
     # HINT: keep documentation sync docs/reference/bazel_macros.rst
 
-    source_config = ":" + ("" if source_dir == "." else source_dir + "/") + "conf.py"
+    config_file_path = join_path(source_dir, "conf.py")
+    sphinx_config_for_bazel_build = ":" + config_file_path
+    has_source_config = len(native.glob([config_file_path], allow_empty = True)) == 1
```
a-zw: Could you use the bundle sources instead of globbing yourself again here?

AlexanderLanin: config.py is not included in the bundle sources

docs.bzl
```diff
@@ -285,14 +341,22 @@ def docs(
...
         srcs = [incremental_src],
         data = docs_data,
         deps = deps,
         env = docs_env
     )
 
+    native.alias(
+        name = "docs",
+        actual = ":_score_docs_cli",
+        tags = ["cli_help=Build documentation:\nbazel run //:docs"],
```
a-zw, line 357:
```suggestion
```

I believe this weird tag is just cargo-culting. I does not do anything, does it?

a-zw (approved):

docs.bzl
AlexanderLanin: we had a cli_help command. It's not cargo cult, as it was invented here. I just don't know if we still have it. Let's call it out of scope :-)

nradakovic (approved):

PR #689, opened 2026-08-04
feat!: needextend effects must be local

AlexanderLanin:
We wanted to ensure local needextend for a while, but it seems we forgot to actually activate the check. Besides activating, this PR also improves and fixes the check itself.

---

Downstream PRs:
* https://github.com/eclipse-score/process_description/pull/759
* https://github.com/eclipse-score/inc_security_crypto/pull/146
* https://github.com/eclipse-score/score/pull/3158


a-zw (approved):
Looks ok to me. No need to fix Copilot's remarks.

We need to fix process though:

> /home/runner/.bazel/sandbox/processwrapper-sandbox/5/execroot/_main/external/score_process+/process/standards/isosae21434/isosae21434.rst:592: WARNING: needextend in S-CORE must always be used per document only. Please add 'c.this_doc()' to the needextend to limit its effects to the correct document. See https://eclipse-score.github.io/docs-as-code/main/how-to/write_docs.html#needextend for more information. [needs.needextend]

a-zw (approved):

AlexanderLanin:
process was fixed in #770, check is green by now

PR #694, opened 2026-08-05
chore: update to process 2.0.3

a-zw:


AlexanderLanin (dismissed):

AlexanderLanin (dismissed):

AlexanderLanin:
## Root cause: integration race between #684 and #676

The `score_process` 2.0.3 update is not the functional cause of this failure. The failure is an integration race between two independently merged changes:

1. #684 changed the `docs()` contract: when `docs/conf.py` is absent, callers must provide both `project` and `project_url`.
2. #676 added `score_docs_compatibility_fixture`, which calls `docs(source_dir = "docs")` without a `docs/conf.py`, `project`, or `project_url`.

Each change can pass CI against its own base: #684 has no such fixture, while #676 was compatible with the previous `docs()` behavior. Once the Merge Queue / synthetic PR merge combines both changes, the fixture violates the new contract. `docs()` fails during analysis before `:docs_bundle` is declared; the later `no such target …:docs_bundle` messages are follow-on errors.

This PR is where the combination became visible because its checks ran against GitHubs synthetic merge ref (`refs/pull/694/merge`) with the current base branch. Thus the queue/merge-result test exposes the race; it is not itself the root cause.

Fix: update `src/tests/docs_bzl/cross_module_fixture/BUILD` to pass `project` and `project_url`, or add `docs/conf.py`.

AlexanderLanin (approved):

PR #701, opened 2026-08-07
docs: cleanup bundles

a-zw:
## 📌 Description

Cleanup our internal use of bundles, so it makes more sense.

URLs stay stable.


BUILD
```diff
@@ -27,24 +27,16 @@ docs(
...
+            "mount_at": "internals/extensions",
         },
         {
             "bundle": "//src/extensions/score_metamodel/docs:metamodel",
             "mount_at": "reference/metamodel",
             "attach_to": "reference/index",
         },
     ],
-    scan_code = [
+    code_targets = [
```
MaximilianSoerenPollak, line 39: This seems a change here, but no change in docs.bzl. Is that correct?

a-zw: Correct. scan_code is deprecated as of 7.0.0 but still working.

MaximilianSoerenPollak (approved):
Seems fine for me, questions answered.

PR #705, opened 2026-08-10
fix: avoid host Git discovery in Bazel sandboxes

4og:
Bazel sandboxed docs actions can walk out of execroot and discover a host Git worktree, causing `ubproject.toml` to be written outside the project. Stop Git-root discovery when the fallback working directory is inside `execroot`, and add a regression test. This intentionally relies on Bazel's execroot layout; it only covers the fallback path, while bazel run continues to use `BUILD_WORKSPACE_DIRECTORY`.


a-zw (approved):

AlexanderLanin:
superseeded by #708

PR #708, opened 2026-08-11
fix: avoid host Git discovery in Bazel sandboxes

AlexanderLanin:
Alternative implementation of #705


4og:
I tested locally with https://github.com/eclipse-score/baselibs/pull/453 and didn't notice any issue

4og (approved):

a-zw (approved):

PR #713, opened 2026-08-12
feat!: dependency upgrades

AlexanderLanin:
and add dependabot config for future updates


.github/dependabot.yml
```diff
@@ -0,0 +1,64 @@
...
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: chore(github-actions)
```
a-zw, line 31:
```suggestion
      prefix: chore(dependabot)
```

a-zw (changes requested):

.github/dependabot.yml
AlexanderLanin: no, the update goes to github-actions. dependabot is only the PR creator, not relevant for the actual change.

a-zw:
Deprecation warning in process:

> WARNING: needs_role_need_template uses the deprecated str.format syntax; migrate '{field}' placeholders to Jinja '{{ field }}' (this will be rendered with str.format for now, but support will be removed in a future release) [needs.deprecated]

Unfortunately, a breaking change for us.

AlexanderLanin:
feat! because of high risk 😆

a-zw (approved):

PR #722, opened 2026-08-17
fix: metrics.json could use more data

### LINKED ISSUE #685: metrics.json could use more data
Carefully select what should be part of that generic json file. I'm sure we need more data there. But not arbitrary data, as it will grow definitely and its very hard to remove anything from it.

Mr-Neutr0n:
Fixes #685

Added a small, documented `needs_overview` (total/external/local/by_type) to the generic metrics.json, updated its JSON schema, added tests, and documented it — the issue's guidance (no arbitrary data, hard to remove fields) was treated as instruction, not followed literally into an arbitrary dump.

Could not run the suite locally: /config/__init__.py:2234. Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/docs/how-to/dashboards_and_quality_gates.rst
+++ b/docs/how-to/dashboards_and_quality_gates.rst
@@ -76,6 +76,22 @@ The documentation build writes ``metrics.json`` via ``score_metrics``, and the `
 
 The dashboard charts and the CI gate both use the same computed metrics.
 
+``metrics.json`` content
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The exported file contains a small, versioned set of fields that is kept
+deliberately stable:
+
+- ``overall_metrics`` – requirement coverage aggregated over all requirement types.
+- ``metrics_by_type`` – the same coverage numbers per requirement type.
+- ``tests`` – testcase linkage statistics, including broken test references.
+- ``needs_overview`` – generic overview of the needs in the metrics scope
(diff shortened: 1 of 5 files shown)


AlexanderLanin:
Hi @Mr-Neutr0n thanks for the contribution! I didn't look deeply into it yet, but doesn't it overlap with `metrics_by_type`?

Mr-Neutr0n:
Thanks for checking. metrics_by_type remains the per-requirement link coverage map, while needs_overview adds total, local versus external, and per-type counts across all needs, including non-requirement types, so the two sections serve different purposes. The current ECA failure is a legal-agreement gate rather than a code or test failure, so it needs to be resolved through the Eclipse contributor process.

MaximilianSoerenPollak:
I think here, we have to figure out what `more data` means firstly before we can go to the implementation phase. 

That is our bad @Mr-Neutr0n , we will clarify this and come back once we have an answer.

Mr-Neutr0n:
Hi @MaximilianSoerenPollak - no rush at all; just checking whether the "more data" definition landed on your side. Happy to pick this back up whenever you have an answer, and fine to close if it is no longer a priority.

FScholPer:
@Mr-Neutr0n please sign the Eclipse ECA in your Eclipse account. Without that we cannot merge your PRs

PR #725, opened 2026-08-18
code: affect optional_link in dec_rec

vibhoraggarwal:
## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/extensions/score_metamodel/metamodel.yaml
+++ b/src/extensions/score_metamodel/metamodel.yaml
@@ -1062,6 +1062,11 @@ needs_extra_links:
     incoming: realized by
     outgoing: realizes
 
+  # Decision Record
+  affects:
+    incoming: affected by
+    outgoing: affects
+
   ##############################################################
   #  S-CORE Metamodel
   ##############################################################
--- a/src/extensions/score_metamodel/tests/rst/options/test_dec_rec.rst
+++ b/src/extensions/score_metamodel/tests/rst/options/test_dec_rec.rst
@@ -1,0 +1,28 @@
+..
+   # *******************************************************************************
+   # Copyright (c) 2026 Contributors to the Eclipse Foundation
+   #
+   # See the NOTICE file(s) distributed with this work for additional
+   # information regarding copyright ownership.
+   #
+   # This program and the accompanying materials are made available under the
+   # terms of the Apache License Version 2.0 which is available at
+   # https://www.apache.org/licenses/LICENSE-2.0
+   #
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+
+.. dec_rec:: Test Decision Record 1
+   :id: dec_rec__test__decision_record_1
+   :status: accepted
+   :context: A decision record is needed for the test.
+   :decision: Use a dedicated decision record need type.
+   :affects: dec_rec__test__decision_record_2
+
+
+.. dec_rec:: Test Decision Record 2
+   :id: dec_rec__test__decision_record_2
+   :status: proposed
+   :context: This decision record is affected by another decision.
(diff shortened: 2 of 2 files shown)


src/extensions/score_metamodel/metamodel.yaml
```diff
@@ -1038,6 +1038,11 @@ needs_extra_links:
     incoming: realized by
     outgoing: realizes
 
+  # Decision Record
+  affects:
```
a-zw, line 1066: Extend the dec_rec type accordingly. I think I actually removed this one because it seemed to be unused.

a-zw (changes requested):

src/extensions/score_metamodel/metamodel.yaml
vibhoraggarwal: it is already added here with dec_rec in the metamodel: https://github.com/eclipse-score/docs-as-code/blob/main/src/extensions/score_metamodel/metamodel.yaml#L990

src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -286,6 +286,12 @@ def test_rst_files(
     all_needs = list(needs_view.values()) + list(extends)
     _validate_need_count(all_needs, rst_data)
 
+    if rst_file == "options/test_dec_rec.rst":
+        needs_by_id = {need["id"]: need for need in needs_view.values()}
+        assert needs_by_id["dec_rec__test__decision_record_1"].get("affects") == [
+            "dec_rec__test__decision_record_2"
+        ]
```
AlexanderLanin: no need to test individual links, they are verified generically. so we dont need to verify "affects" with a custom test here.

src/extensions/score_metamodel/tests/rst/options/test_dec_rec.rst
```diff
@@ -0,0 +1,28 @@
...
+   # SPDX-License-Identifier: Apache-2.0
+   # *******************************************************************************
+
+
+.. dec_rec:: Test Decision Record 1
+   :id: dec_rec__test__decision_record_1
+   :status: accepted
+   :context: A decision record is needed for the test.
+   :decision: Use a dedicated decision record need type.
+   :affects: dec_rec__test__decision_record_2
```
AlexanderLanin, line 21: nitpick: just for sanity. DR2 affects DR1, not the other way around?! 😆

AlexanderLanin:
"affects" seems like such a trivial word... but its not. People will interpret it differently. Is there any description anywhere on where we would expect people to use `affects`?

src/extensions/score_metamodel/tests/test_rules_file_based.py
vibhoraggarwal: removed the explicit test.

vibhoraggarwal:
> "affects" seems like such a trivial word... but its not. People will interpret it differently. Is there any description anywhere on where we would expect people to use `affects`?

This repo does not host the definition of affects but only mention in [metamodel.yaml](https://github.com/eclipse-score/docs-as-code/blob/3c02b6b449348ca3b4114ab4020ab9dc0cffc5ee/src/extensions/score_metamodel/metamodel.yaml#L1014). I also browsed in the [process_description ](https://github.com/search?q=repo%3Aeclipse-score%2Fprocess_description+affects&type=code)which does not provide it either. I did not want to scope this PR into describing `affects` though.

AlexanderLanin (approved):

src/extensions/score_metamodel/metamodel.yaml
a-zw: You are right.

a-zw (approved):

PR #730, opened 2026-08-19
refactor(module-verification-report): split, tests, docs

antonkri:
This PR introduces an extension to generate a module verification report, that can be used by every module. For detailed documentation see module_verification_report.rst

See https://github.com/eclipse-score/baselibs/pull/491#issuecomment-5343947633 for an example usage

---

**Note:** the docs_and_test Bazel macro (chaining test/coverage execution with the docs build) has been split out into a separate PR: #759. This PR now covers only the score_module_verification_report Sphinx extension itself.

**Depends on:** #739 — testcase result annotation (the colored (passed)/(failed) badge on testcase back-links) has been extracted, generalized, and moved to score_source_code_linker there. This PR relies on that hook to decorate the testcase back-links rendered inside the module verification report; please merge #739 first (or review it alongside this PR).


src/extensions/docs/module_verification_report.rst
```diff
@@ -0,0 +1,447 @@
...
+   * - Aspect
+     - Contract
+
+   * - Directive
+     - ``.. module-verification-report::`` — no arguments, no content;
+       one optional ``:config:`` option.
+
+   * - Config option
+     - ``:config: <path>`` — YAML file resolved relative to Sphinx's
+       ``srcdir`` (i.e. the directory containing ``conf.py``).
```
a-zw: Why a YAML file? Could we configure it via directive attributes instead?

src/extensions/docs/module_verification_report.rst
```diff
@@ -0,0 +1,447 @@
...
+
+   * - Reads from the source tree
+     - Every ``.. mod::`` and ``.. comp::`` need, discovered by a
+       shallow regex scan at ``env-before-read-docs`` time.
```
a-zw: Reading the source tree is an anti-pattern. Could use a needs filter instead?

src/extensions/score_module_verification_report/scanner.py
```diff
@@ -0,0 +1,184 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Filesystem scanner for ``.. mod::`` / ``.. comp::`` needs.
+
+We deliberately do **not** query ``SphinxNeedsData`` here: doing so would
+require reading the report source strictly after every source registering
+a ``.. comp::`` / ``.. document::`` need, which forces
+``parallel_read_safe = False`` on the extension and produces two
+Sphinx-level warnings per build (``the score_module_verification_report
+extension is not safe for parallel reading`` / ``doing serial read``).
```
AlexanderLanin: looking through the PR, this is the most obvious red flag that we need to investigate. scanning source code is less than ideal.

src/extensions/docs/module_verification_report.rst
antonkri: Good hint, adapted!

src/extensions/docs/module_verification_report.rst
antonkri: Done

src/extensions/score_module_verification_report/scanner.py
antonkri: Done!

src/extensions/score_module_verification_report/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+
+Implementation is split across:
+
+* :mod:`.templates`   — RST templates + default workproduct lists + CSS
+* :mod:`.rendering`   — template expansion / report body assembly
+* :mod:`.directive`   — the ``ModuleVerificationReportDirective`` class
+* :mod:`.testcase_annotations` — ``doctree-resolved`` badge decoration
+  for ``testcase__…`` back-links on pages that render the directive
+* :mod:`.consistency_checks` — ``build-finished`` validation that every
+  component is properly linked in the needs graph
```
a-zw: Why are consistency checks not handled by the score_metamodel extension?

src/extensions/score_module_verification_report/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    .. module-verification-report::
+       :module-id: mod__baselibs
+       :feature-id: feat__baselibs
+       :components: comp__baselibs_json,
+                    comp__baselibs_bit_manipulation,
+                    comp__baselibs_containers
+
+An optional ``:config:`` YAML file can supply non-default workproducts or
+per-component doc-id overrides for the rare case where component documents
+do not follow the standard naming convention.
```
MaximilianSoerenPollak: This is a user error. 
This should be fixed `upstream` instead of adding unnecessary complexity in the code

src/extensions/score_module_verification_report/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+def setup(app: Any) -> dict:
+    app.add_directive("module-verification-report", ModuleVerificationReportDirective)
+    app.connect("env-before-read-docs", init_docnames)
+    app.connect("env-before-read-docs", init_registry)
+    app.connect("env-purge-doc", purge_docname)
+    app.connect("env-purge-doc", purge_registry)
+    app.connect("env-merge-info", merge_docnames)
+    app.connect("env-merge-info", merge_registry)
+    app.connect("doctree-resolved", annotate_testcase_results)
+    app.connect("build-finished", check_consistency)
```
a-zw: I like the idea of this directive but all these connect calls are suspicious to me. Is this complexity really necessary?

src/extensions/score_module_verification_report/templates.py
```diff
@@ -0,0 +1,324 @@
+# *******************************************************************************
```
MaximilianSoerenPollak: This looks very close / similar t a manual jinja2 template that sphinx-needs natively already supports. 

Is there any pros and cons to using this approach in comparison to jinja2. 
Would using jinja2 remove some of this complexity, or just introduce different ones?

src/extensions/score_module_verification_report/consistency_checks.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if comp_need is None:
+            logger.warning(
+                "[module-verification-report] %s: "
+                "%s (listed in :components:) not found in "
+                "sphinx-needs — check the id",
+                docname,
+                comp_id,
+            )
+            continue
```
MaximilianSoerenPollak: Components are already in the `module` therefore no need to `re-iterate` them in the mod_verification_report if you already link the module via `:module-id:`

src/extensions/score_module_verification_report/consistency_checks.py
```diff
@@ -0,0 +1,155 @@
...
+        if mod_need is not None:  # noqa: SIM102
+            if comp_id not in mod_need.get("includes", []):
+                logger.warning(
+                    "[module-verification-report] %s: "
+                    "%s is listed in :components: but not in "
+                    "%s :includes:",
+                    docname,
+                    comp_id,
+                    module_id,
+                )
```
MaximilianSoerenPollak: Same comment above.

src/extensions/score_module_verification_report/directive.py
```diff
@@ -0,0 +1,144 @@
...
+        if not comp_id:
+            continue
+        slug = (
+            comp_id[len(component_prefix) :]
+            if component_prefix and comp_id.startswith(component_prefix)
+            else comp_id
+        )
+        title = slug.replace("_", " ").title()
+        result.append({"id": comp_id, "slug": slug, "title": title})
+    return result
```
MaximilianSoerenPollak: I do not understand why we remove the `version` here? 

Is this because it's a `sphinx directive` instead of a `sphinx-needs directive`?

src/extensions/score_module_verification_report/__init__.py
a-zw: There already is `mod_ver_report` in the metamodel. This should probably be added into the template here for process compliance.

Also, we can extend `mod_ver_report` and then leave the consistency checking to existing mechanisms in the score_metamodel extension.

a-zw (changes requested):

src/extensions/score_module_verification_report/testcase_annotations.py
```diff
@@ -0,0 +1,107 @@
+# *******************************************************************************
```
MaximilianSoerenPollak: This is a great idea. 

However I think we should pull this out of this extension and put it into the test-case linking so it is done for `all` testcaseneeds instead of just for the ones inside this verification report.

src/extensions/score_metamodel/external_needs.py
```diff
@@ -190,7 +190,7 @@ def add_external_needs_json(e: ExternalNeedsSource, config: Config):
     assert isinstance(config.needs_external_needs, list)  # pyright: ignore[reportUnknownMemberType]
     config.needs_external_needs.append(  # pyright: ignore[reportUnknownMemberType]
         {
-            "base_url": needs_json_data["project_url"]
+            "base_url": needs_json_data["project_url"].rstrip("/")
```
MaximilianSoerenPollak: Not quiet sure why this was changed, any reason for it?

MaximilianSoerenPollak (commented):
Thanks for the work, some comments & questions from my side here.

src/extensions/score_module_verification_report/__init__.py
antonkri: Hey, yes, unfortunately. I've reduced the number of app.connects a little bit, but most of them should stay for two reasons:

1. We execute check_consistency only after the whole build was done. For this we need notification. Additionally, in the check_consistency function we need information from the module_verification_report directive, which components were configured. After the build is finished, this information is not available anymore as all objects were destroyed. Therefore we need the registry.

2. We want to append the test result to the test names, and this can happen only after doctree was resolved, otherwise we can miss some of the tests. For this we need doctree-resolved notification.

src/extensions/score_module_verification_report/__init__.py
antonkri: Right, I've extended code and configuration to generate the mod_verification_report, for now only with mandatory arguments.

src/extensions/score_module_verification_report/__init__.py
antonkri: Right, that's legacy comment, removed it.

src/extensions/score_module_verification_report/templates.py
antonkri: Yes, for now the templates are very simple. I am not sure, if jinja wouldn't be an overkill here. We could think about this in the future.

src/extensions/score_module_verification_report/consistency_checks.py
antonkri: That's done for the following reason. Current implementation relies on the fact, that the component ids and some additional information is known before the sphinx-build is started, e.g. I can filter for the requirements of the component only if I know component id. Trying to resolve all component ids belonging to one module and building up the filters for these components in parallel as part of one sphinx-build can lead to raise conditions (may be there is a good way with sphinx how to do this, I did not manage to find it out). Therefore I specify a minimal amount information as config arguments of module_verification_report directive. Afterwards, as build has finished, I check for the consistency to ensure, that configuration attributes were correct.

src/extensions/score_module_verification_report/consistency_checks.py
antonkri: same comment as above

src/extensions/score_module_verification_report/directive.py
antonkri: yes, in the configuration file we just specify for which components we want to generate the module verification report, not in which version. As someone can accidentally provider version number, we want to ignore it for now.

src/extensions/score_module_verification_report/testcase_annotations.py
antonkri: I think Alexander has already did it.

src/extensions/score_metamodel/external_needs.py
antonkri: by mistake, this change was abonden

src/extensions/score_module_verification_report/consistency_checks.py
AlexanderLanin: we need to figure out a way to move this PR to a needs based setup. The easiest way would be by splitting it up. not sure how yet.

bzl/docs_and_test.bzl
```diff
@@ -0,0 +1,124 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Bazel macro that chains ``bazel test`` (or ``bazel coverage``) with a
+docs target in one command.
```
AlexanderLanin: this should be a separate pull request. I dont understand why this works, while normally we cannot run bazel from bazel, but that can be discussed there.

src/extensions/score_module_verification_report/directive.py
```diff
@@ -0,0 +1,198 @@
...
+    The id follows the ``<Req Type>__<Abbreviations>__<Architectural
+    Element>`` scheme mandated for 3-part need types (``mod_ver_report``
+    declares ``parts: 3`` in metamodel.yaml), so exactly two ``__``
+    separators are required: ``mod_vrep__<module_short>__report``.
+    """
+    title_case = module_short.replace("_", " ").title()
+    return f"mod_vrep__{module_short}__report", f"{title_case} Verification Report"
+
+
+class ModuleVerificationReportDirective(SphinxDirective):
```
AlexanderLanin, line 83: did you try using a simple needs template instead of a directive?

AlexanderLanin: while that could simplify the technical side significantly (if it works), it doesnt look good in the generated docu

AlexanderLanin: on 3rd thought that could simplify the PR 10x, I need to try it

AlexanderLanin: left side is 10x less code:

image

AlexanderLanin: discussion in slack how to continue

AlexanderLanin:
moving on to #763

PR #739, opened 2026-08-21
feat(testcase): annotate testcase links with results

AlexanderLanin:
## Summary

* move testcase result annotations into `score_source_code_linker`
* decorate testcase backlinks with passed, failed, skipped, disabled, or fallback statuses

Extracted from #730; original author is @antonkri 

---

see https://eclipse-score.github.io/docs-as-code/pr-739/internals/requirements/requirements.html
image


src/extensions/score_source_code_linker/tests/test_source_code_link_integration.py
```diff
@@ -250,8 +250,25 @@ def basic_conf():
...
+    "testlink",
+    "name",
+    "file",
+    "line",
+    "test_type",
+    "derivation_technique",
+    "result",
+    "result_text",
 ]
-needs_extra_options = ["source_code_link", "testlink"]
```
MaximilianSoerenPollak: These are options though, not links?

src/extensions/score_source_code_linker/testcase_annotations.py
```diff
@@ -0,0 +1,169 @@
...
+def _testcases_by_external_url(needs: Any) -> dict[str, list[dict[str, Any]]]:
+    """Group testcase needs by the URL used for their rendered GitHub link."""
+    testcases_by_url: dict[str, list[dict[str, Any]]] = {}
+    for need in needs.values():
+        if need.get("type") != "testcase":
+            continue
+        external_url = need.get("external_url")
+        if external_url:
+            testcases_by_url.setdefault(external_url, []).append(need)
+    return testcases_by_url
```
MaximilianSoerenPollak: Why?

MaximilianSoerenPollak (dismissed):
Overall looks okay I think. 

It works in the implementation and rendered page and looks good.

src/extensions/score_source_code_linker/testcase_annotations.py
AlexanderLanin: The URL is built from the repository, file, and source line, so it identifies a source location, not necessarily one testcase. Parameterized tests or multiple testcases reported for the same line can share it; the testcase IDs are more specific because they also contain the testcase name.

The previous implementation stored one need per URL, so a later need silently replaced an earlier one and could make a link show the wrong result. The list preserves all candidates. References with a resolved refid use the exact testcase; URL-only references are annotated only when exactly one candidate exists. Otherwise we leave the annotation off rather than showing an arbitrary result.

I added this rationale and the related behavior documentation in commit eda4e2cb, which is now pushed to the PR branch.

src/extensions/score_source_code_linker/testcase_annotations.py
```diff
@@ -0,0 +1,211 @@
...
+    if refuri:
+        candidates = testcases_by_url.get(refuri, [])
+        if len(candidates) == 1:
+            return candidates[0]
+
+    return None
+
+
+def annotate_testcase_results(app, doctree, docname):
+    """Append a coloured, theme-compatible result annotation to testcase refs.
```
a-zw: How could this extension know which theme is used? This is not "theme-compatible" it is "hardcoded for the theme currently used in S-CORE". At least let's be honest about this junk.

src/extensions/score_source_code_linker/tests/test_testcase_annotations.py
```diff
@@ -0,0 +1,185 @@
...
+    ref = nodes.reference("", "", nodes.Text(text), **attrs)
+    doc.append(ref)
+    return doc, ref
+
+
+def _app(needs):
+    return SimpleNamespace(env=SimpleNamespace(needs_view=_FakeNeedsView(needs)))
+
+
+def test_annotates_external_github_testlink():
```
a-zw, line 42: Create a requirement and link these tests to it.

docs/how-to/test_to_doc_links.rst
```diff
@@ -80,35 +93,21 @@ Then as described above, each testcase also has properties inside the XML.
...
-
-Each language uses a different mechanism to attach properties to test cases,
-but all produce the same JUnit XML output that the linker consumes.
-
-C++ (gTest)
-   Use ``RecordProperty`` — shared properties go in ``SetUp()``, per-test properties
-   inside each ``TEST_F``.
-   See :need:`gd_req__verification_link_tests_cpp`.
-
-Rust
```
a-zw, line 103: Rust got lost. We should mention it explicitly.

docs/how-to/test_to_doc_links.rst
```diff
@@ -20,21 +20,35 @@ Reference Docs in Tests
...
+
+See the `Verification Templates <https://eclipse-score.github.io/reference_integration/main/_collections/score_process/process/process_areas/verification/guidance/verification_templates.html>`_
+for complete examples and the required metadata.
 
+Advanced usage: JUnit XML for other languages
+----------------------------------------------
 
-Required Properties
--------------------
+This section is only relevant when your language or test framework does not
```
a-zw, line 50: ASD-STE100 Simplified Technical English is a worthwhile guideline when writing how-tos.

a-zw (dismissed):

.github/workflows/license_check.yml
```diff
@@ -25,9 +25,18 @@ permissions:
...
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          persist-credentials: false
+
+      - uses: eclipse-score/dash-license-scan@fe2f9ab2e0cd2bfac7efdbe2637e188b1a5e2025 # dev as of 2026-08-21
+        with:
+          token: ${{ secrets.ECLIPSE_GITLAB_API_TOKEN }}
+          eclipse_project: ${{ vars.ECLIPSE_PROJECT }}
+          lockfiles: src/requirements.txt
+          trigger_review: false
+          comply_with: ASF EF
```
MaximilianSoerenPollak: Why is this in this PR? Seems unrelated

docs/how-to/test_to_doc_links.rst
MaximilianSoerenPollak: I agree, we should mention that currently there is no way to do this in rust.

MaximilianSoerenPollak (dismissed):
Seems okay though still some questions open that can be adressed in future PRs. 

Would like @a-zw approval too before merge.

src/extensions/score_source_code_linker/testcase_annotations.py
AlexanderLanin: yeah its theme compatible to the current theme, I'll improve wording

src/extensions/score_source_code_linker/tests/test_testcase_annotations.py
AlexanderLanin: need to discuss whether this gets requirements or not, since most of our code does not have requirements. either none, specific or something in the middle like a "ux requirement".

docs/how-to/test_to_doc_links.rst
AlexanderLanin: this talks about programming language

.github/workflows/license_check.yml
AlexanderLanin: yeah it is, just an attempt to change it. if it doesnt work out, we should drop the license check completely

AlexanderLanin: removed from PR

docs/how-to/test_to_doc_links.rst
AlexanderLanin: documented

src/extensions/score_source_code_linker/tests/test_testcase_annotations.py
AlexanderLanin: linked to testcase requirement

src/extensions/score_source_code_linker/testcase_annotations.py
AlexanderLanin: wording improved

docs/how-to/test_to_doc_links.rst
MaximilianSoerenPollak: Might be a good change, but not in the scope of this PR.

src/extensions/score_source_code_linker/testcase_annotations.py
```diff
@@ -0,0 +1,211 @@
...
+_FALLBACK_CLASS = "score-testcase-result--unknown"
+# The event handler is expected to be idempotent for a doctree. This marker
+# prevents a second invocation from appending the same status again.
+_ANNOTATED_ATTR = "score_source_code_linker_testcase_result_annotated"
+
+# The styles are inserted into a document only when that document contains an
+# annotation. Keeping them in one block avoids repeating inline style rules on
+# every testcase link and lets the same classes handle theme changes.
+_TESTCASE_STATUS_CSS = """
+<style>
```
MaximilianSoerenPollak, line 42: for future PR maybe move this to overall css files

MaximilianSoerenPollak (changes requested):
Talked about in a call, can be simplified a lot. 
Work in Progress.

MaximilianSoerenPollak (approved):
Talked about in a call. Looks much simpler now. 💯

PR #741, opened 2026-08-21
feat: derive required_in_id for generated conf

AlexanderLanin:
- derive `required_in_id` from Bazel's `native.module_name()` for generated `conf.py` files
- strip the first underscore-separated prefix (`score_docs_as_code` → `docs_as_code`)


MaximilianSoerenPollak (approved):
Talked over in a call.

PR #742, opened 2026-08-21
fix: docs_bundle with local puml files

mariuswbr:
## 📌 Description
`_canonical_mount_dir` incorrectly skipped symlink resolution for in-tree source bundles, causing a false-positive ExtensionError from sphinx-mounts' path confinement check when building in a Bazel sandbox. The fix removes the early-return guard so the file-following logic runs for all bundle types. Two unit tests are added covering the in-tree sandbox case and the empty-bundle fallback.

## 🚨 Impact Analysis

## ✅ Checklist


MaximilianSoerenPollak:
Can you point to where this actually breaks in your module? I'm a bit curious about these symlinked files how they look / etc. 
Like the structure / tree of the folders etc.

a-zw (approved):

PR #758, opened 2026-08-25
fix: avoid duplicate needs from nested docs bundles

AlexanderLanin:
## Summary

- expose only direct bundle sources as Sphinx top-level sources while keeping transitive files available for mounts
- add producer/consumer docs_bzl coverage for a docs_bundle in a subdirectory
- verify public docs() targets and prevent duplicate needs in needs_json

## Bug

When a docs() target included a docs_bundle from a subdirectory, the complete transitive bundle DefaultInfo was passed as Sphinx srcs. Nested sources were then processed once as raw sources and once through score_mounts, causing needs.duplicate_id failures when the docs bundle was consumed by another docs() target.

Required for https://github.com/eclipse-score/score/pull/3215

## Validation

- targeted subdirectory_bundle scenario: 3 passed
- public docs() target conflict check: passed
- existing nested, external, and basic docs_bzl tests: passed
- sandboxed data-bundle needs_json build: passed
- commit hooks: Ruff, BasedPyright, and copyright checks passed

Note: the existing data_files_runfiles runtime smoke test can encounter workspace-wide test-log warnings in a dirty workspace; its needs_json build passes.


MaximilianSoerenPollak (dismissed):
Looked through in a call. Looks alright 💯

MaximilianSoerenPollak (approved):
Talked about in a call. Looks good 💯

PR #768, opened 2026-08-27
test: ignore downstream Bazel lockfiles

AlexanderLanin:
## Summary

We don't care about lock file problems in downstream repos, however they lead to problems in our downstream tests.

- explicitly run downstream Bazel commands with `--lockfile_mode=ignore`
- prevent consumer repository `.bazelrc` settings such as `common --lockfile_mode=error` from blocking compatibility tests


MaximilianSoerenPollak (dismissed):
Will make things easier.

MaximilianSoerenPollak (approved):

PR #763, opened 2026-08-26
feat: module verification report / complex templates

AlexanderLanin:
* Module Verification Report content/template by @antonkri, see #730
* Idea to use needtemplates by @MaximilianSoerenPollak
* Telling AI to combine the two ideas by @AlexanderLanin 🤣
* Crazy analysis in #764

PR #769, opened 2026-08-27
feat: add L+ Jinja report materialization

AlexanderLanin:
## Summary

This draft PR adds the L+ proof of concept for model-complete Sphinx-Needs reports.

L+ keeps ordinary documentation reading parallel. During the read phase it records a report placeholder. After all worker environments have been merged, it renders the selected Jinja template against the resolved Needs model and materializes the resulting report in the existing document. The report outline is derived from Need links, and the template can use existing Sphinx-Needs directives such as `needtable` and `needpie`.

The previous manifest/generator experiment is intentionally not included. The report definition lives in `src/needs_templates/mod_ver_report_tiny.need`.

## Evaluation conclusion

L+ is technically valid as a controlled solution for this report, but it is not a lightweight general-purpose extension. Sphinx does not provide a public API for adding model-complete sections and local table-of-contents entries after parallel Need collection has finished. L+ therefore needs a small compatibility layer around the cached doctree and the local ToC collector.

This makes L+ a reasonable prototype when automatic discovery from the Need graph is mandatory. It is not the simplest long-term option. If an explicit, reviewable component list is acceptable, the K approach from PR #730 is smaller, uses normal read-phase Sphinx sections and public APIs, and should be preferred.

The implementation should therefore be reviewed as an L+ experiment with a strict acceptance gate, not as proof that late materialization is a general replacement for ordinary Sphinx source processing.

## Validation

- `bazel test //src/extensions/score_sphinx_needs_templates:report_pipeline_tests`
- `bazel run //:docs`

### DIFF
--- a/src/extensions/score_sphinx_bundle/BUILD
+++ b/src/extensions/score_sphinx_bundle/BUILD
@@ -36,6 +36,7 @@ py_library(
         "@score_docs_as_code//src/extensions/score_metamodel",
         "@score_docs_as_code//src/extensions/score_mounts",
         "@score_docs_as_code//src/extensions/score_source_code_linker",
+        "@score_docs_as_code//src/extensions/score_sphinx_needs_templates",
         "@score_docs_as_code//src/extensions/score_metrics",
         "@score_docs_as_code//src/extensions/score_sync_toml",
         "@score_docs_as_code//src/helper_lib",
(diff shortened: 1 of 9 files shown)


AlexanderLanin:
moving on to #763

PR #773, opened 2026-08-30
feat(module-verification-report): show test results

antonkri:
The requirements tables of the module verification report listed the backlinks `fully_verifies_back` and `partially_verifies_back`, which render as bare need ids (e.g. `testcase__CheckBit__WithUInt8_gavap`). That neither tells the reader whether a test passed nor links to the test source.

Use the `testlink` option instead, which is populated by the `test_code_linker` string link. It renders the test name together with its result and links to the corresponding line in the test file. The `doctree-resolved` hook `annotate_testcase_results` then colours the `(passed|failed|skipped|disabled)` suffix.

Applied to both requirements tables:
- feature level ("Requirements Statistics", feat_req)
- component level ("Requirements Traceability", comp_req)

Verified against eclipse-score/baselibs: the report now renders 43 `(passed)` badges with clickable source links, up from none.

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/needs_templates/module_verification_report.need
+++ b/src/needs_templates/module_verification_report.need
@@ -140,8 +140,8 @@ Requirements Statistics
    .. needtable::
       :filter: type == "feat_req" and "{{ feature_id }}" in satisfied_by
       :style: table
-      :columns: id;title;safety;status;fully_verifies_back;partially_verifies_back
-      :colwidths: 13,22,8,10,23,24
+      :columns: id;title;safety;status;testlink
+      :colwidths: 13,22,8,10,47
       :sort: id
 
 Architecture Statistics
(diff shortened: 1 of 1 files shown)


src/needs_templates/module_verification_report.need
```diff
@@ -140,8 +140,8 @@ Requirements Statistics
    .. needtable::
       :filter: type == "feat_req" and "{{ feature_id }}" in satisfied_by
       :style: table
-      :columns: id;title;safety;status;fully_verifies_back;partially_verifies_back
-      :colwidths: 13,22,8,10,23,24
+      :columns: id;title;safety;status;testlink
```
MaximilianSoerenPollak, line 143: According to the title you want to show test results. 
This will not just show test results, it will show the `testlink` of the requirement which will give you a link with the result tacked on at the end. 
If that is what you want then this is correct, if you `just` want the results, you have to do this differently and most likely write some python code that gives you this in the return.

AlexanderLanin:
merging as-is for now, since initial template was also not reviewed. will address in next process meeting

AlexanderLanin (approved):

PR #774, opened 2026-08-30
fix: treat '<error>' testcases as failed instead of passed

antonkri:
`parse_testcase_result` only inspected the `<failure>` and `<skipped>` child elements of a `<testcase>`:

    if skipped is None and failed is None:
        return "passed", ""

JUnit XML also uses `<error>` for testcases that did not complete, e.g. an uncaught exception in Python's `unittest` or a crashing binary. Such a testcase has neither `<failure>` nor `<skipped>`, so it fell into the branch above and was reported as `passed`.

That is the worst possible failure mode for a verification report: a broken test silently shows up as a green `(passed)` badge next to the requirement it is supposed to verify.

Handle `<error>` explicitly and map it to `failed`, using the element's `message` attribute as the result text. An existing `<failure>` still takes precedence, so the reported message stays the assertion message when both elements are present.

## 🚨 Impact Analysis

## ✅ Checklist

### DIFF
--- a/src/extensions/score_source_code_linker/tests/test_xml_parser.py
+++ b/src/extensions/score_source_code_linker/tests/test_xml_parser.py
@@ -247,6 +247,17 @@ def test_parse_testcase_result():
     ET.SubElement(tc4, "skipped", {"message": "skp"})
     assert xml_parser.parse_testcase_result(tc4) == ("skipped", "skp")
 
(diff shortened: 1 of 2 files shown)


MaximilianSoerenPollak:
Do you have an `error` XML example? I have never seen that and I'm quiet curious how it would look / when it happens.

AlexanderLanin (approved):
LGTM, lets wait for copilot

PR #779, opened 2026-08-31
fix: preserve per-entry bundle data when rebasing

### LINKED ISSUE #777: docs_bundle propagates transitive data to unrelated mount entries
## Problem

When a project consumes an external `docs_bundle` with nested data-only bundles, the generated data files are propagated to every visible bundle entry instead of only to the entry that owns them. The resulting manifest contains the same data files on the normal source mount and on the data-only mount.

`score_mounts._resolve_data_mounts()` then deduplicates by the physical parent directory and keeps the first `MountSpec`. This assigns generated files to the wrong `mount_at` and produces `sphinx-mounts: docname conflict` warnings; the intended generated documentation is skipped.

## Reproduction

Using `score_docs_as_code` 8.1.0, consume `@score_docs_as_code//:docs_bundle` from `eclipse-score/reference_integration` with a placement such as:

```python
{
    "bundle": "@score_docs_as_code//:docs_bundle",
    "mount_at": "process_methods_tools/docs_as_code",
    "attach_to": "process_methods_tools",
}
```

The generated manifest contains the metamodel data on the root docs entry as well as on the `reference/metamodel` data entry. The same happens for the nested bundle test data.

Observed result:

```text
sphinx-mounts: docname conflict for process_methods_tools/docs_as_code/index
```

The generated metamodel file contains the referenced label, but its mount is skipped because of this conflict.

## Expected behavior

Each manifest entry should carry only the data belonging to that entry, or the data resolver should retain the mount metadata associated with each data bundle. A consuming project should be able to mount a composed external bundle without data mounts being redirected to the first source entry.

## Context

This was found while integrating `score_docs_as_code` 8.1.0 into `eclipse-score/reference_integration`. The synthetic `__data__/...` runtime paths introduced in 8.1.0 distinguish data-only bundles, but the data metadata propagation problem remains.

AlexanderLanin:
Fixes #777

## Change

A bundle data set is metadata of an individual mount entry, not of every entry reachable through that bundle.

- _rebase_bundle_entry() now retains entry.data while changing only the entry placement metadata (mount_at and attach_to).
- Its caller no longer supplies child DocsBundleInfo data, which aggregates data from the child and all of its nested bundles.

This prevents a generated file owned by one source or data-only entry from being repeated on unrelated entries in the generated mounts manifest. The data resolver can consequently retain the mount that owns the file.

## Regression coverage

The nested-bundles consumer scenario mounts a parent bundle that owns generated_output.txt and contains a child bundle. The test asserts that the parent manifest entry contains that file and that the child entry has an empty data list. The latter assertion fails with the previous aggregate-data propagation.

Verified with:

- .venv_docs/bin/python -m pytest -vv src/tests/docs_bzl/test_nested_bundles.py::test_nested_bundles_render_and_preserve_metadata


MaximilianSoerenPollak (approved):
Seems plausible. Have to test in ref-int if this actually fixes it.

PR #782, opened 2026-08-31
fix: respect Bazel package boundaries in mounted docs

### LINKED ISSUE #781: fix: respect Bazel package boundaries when mounting docs bundles
## Problem

`docs_bundle(source_dir = "docs")` discovers its own sources with Bazel `glob()`. Bazel excludes subdirectories that are separate packages because they contain a `BUILD` file.

The generated mount manifest, however, contains only the physical `source_dir`. At runtime, `sphinx-mounts` walks that directory recursively and does not know Bazel package boundaries. If a child package below `docs/` is mounted as its own bundle, its files are mounted twice:

- once through the parent directory mount;
- once through the explicit child bundle mount.

This produces `sphinx-mounts: docname conflict` warnings during `bazel run //:docs`.

## Scope: both supported composition levels

This is not limited to a bundle mounted directly by `docs()`. The same failure occurs at both levels whenever the child bundle source directory is physically below the parent source directory:

1. **A `docs_bundle` inside `docs()`:** `docs()` creates a primary `docs_bundle`, while Sphinx walks the live `source_dir` and `score_mounts` mounts the child bundle. The child documentation can therefore be discovered once by Sphinx through the primary tree and once by the explicit mount.
2. **A `docs_bundle` inside another `docs_bundle`:** the parent bundle is mounted in directory mode and recursively walks its source directory, while the child bundle is mounted as a separate manifest entry. The parent mount discovers the child documentation and the child mount discovers it again.

Logical nesting with physically disjoint source directories does not trigger this particular collision. The problem is the mismatch between Bazel package-aware source ownership and recursive directory discovery.

## Expected behavior

The runtime must expose exactly the source files selected by each owning `docs_bundle`, at every composition level:

- the primary Sphinx source tree must not discover files owned by nested packages;
- a parent directory mount must not discover files owned by a nested child bundle; and
- the explicit child mount must remain responsible for the child files.

This must hold for both `bazel run ...:docs` and sandboxed `needs_json` builds, without duplicate documents or Need IDs.

## Reproduction: child bundle mounted by `docs()`

```
docs/
  index.rst
  features/
    foo/
      BUILD
      index.rst
```

Parent:

```starlark
docs(
    source_dir = "docs",
    bundles = [{
        "bundle": "//docs/features/foo:docs",
        "mount_at": "features/foo",
    }]
)
```

Child:

```starlark
docs_bundle(name = "docs", source_dir = ".")
```

Bazel's parent `glob("docs/**/*")` excludes `docs/features/foo/**`, but the current mount rooted at `docs/` includes it anyway.

## Reproduction: child bundle inside another `docs_bundle`

```
parent-docs/
  index.rst
  child/
    BUILD
    index.rst
```

Parent bundle:

```starlark
docs_bundle(
    name = "parent",
    source_dir = "parent-docs",
    bundles = [{
        "bundle": "//child:docs",
        "mount_at": "child",
    }]
)
```

Child bundle:

```starlark
docs_bundle(name = "docs", source_dir = ".")
```

When `parent` is mounted by a consumer, its directory entry rooted at `parent-docs/` recursively discovers `child/index.rst`, and the explicit child entry mounts the same document again.

## Proposed direction

Extend the bundle mount metadata with each source entry's selected file list (or excluded nested package roots). The runtime must honor that selection for every directory-mode bundle mount and also constrain primary Sphinx discovery for the `docs()` source tree. Add integration scenarios for both composition levels and assert that `bazel run ...:docs` and `needs_json` succeed without duplicate documents.

## Related

- https://github.com/eclipse-score/docs-as-code/pull/758 fixes duplicate *sandboxed Sphinx inputs* from nested bundles, but its scenario places the child source outside the parent `source_dir`.
- eclipse-score/score#3215 exercises the missing runtime-mount case.

AlexanderLanin:
## Summary

Make runtime Sphinx discovery follow the same direct source ownership that Bazel applies to `docs_bundle(source_dir = ...)`.

The mounts manifest now records an anchored include allowlist for each bundle and a separate `primary_source` record for the project source tree. `score_mounts` passes the allowlist to sphinx-mounts for child bundles and excludes unowned RST/Markdown files from the primary Sphinx source tree.

This prevents a parent source directory from recursively exposing documents below nested Bazel packages. Those documents are then supplied exactly once by their declared child bundle.

## Regression coverage

The `subdirectory_bundle` scenario now places the child bundle under the producer's `docs/` directory, with its own `BUILD` file. It verifies producer and consumer `needs_json`, rendered docs, strict docs checks, and link checks. Resolver coverage verifies the new manifest fields; the existing nested-bundles scenario remains unchanged and passes.

Fixes #781

### DIFF
--- a/bzl/bundle_rules.bzl
+++ b/bzl/bundle_rules.bzl
@@ -62,6 +62,7 @@ DocsBundleInfo = provider(
     doc = "A documentation bundle with its source and placement metadata.",
     fields = {
         "entries": "Ordered entries, one per source directory, including its final documentation-tree location.",
+        "own_source_entry": "This bundle's own source directory entry, or None for an aggregator.",
         "own_source_files": "This bundle's direct source files, excluding nested bundles.",
         "sourcelinks": "Source-code-link JSON files together with their owning repository.",
         "external_runfiles": "Documentation source files from external repositories needed in runfiles.",
@@ -197,6 +198,7 @@ def _rebase_bundle_entry(entry, mount_at, attach_to):
         entry_doc = entry.entry_doc,
         external = entry.external,
         repository = entry.repository,
+        source_files = entry.source_files,
         data = entry.data,
     )
 
@@ -244,11 +246,12 @@ def _docs_bundle_impl(ctx):
     own_source_files = []
     own_external_runfiles = []
     own_data = depset(direct = ctx.files.data)
+    own_source_entry = None
 
     if ctx.files.srcs:
         runtime_path = _bundle_runtime_path(ctx)
         external = runtime_path.startswith("../")
-        entries.append(struct(
+        own_source_entry = struct(
             runtime_path = runtime_path,
             # The execution root and runfiles tree spell external repositories
             # differently. Keep both locations so every public docs() target can
@@ -259,8 +262,10 @@ def _docs_bundle_impl(ctx):
             entry_doc = ctx.attr.entry_doc,
             external = external,
             repository = ctx.label.workspace_name,
+            source_files = ctx.files.srcs,
             data = own_data,
-        ))
+        )
+        entries.append(own_source_entry)
         own_source_files.extend(ctx.files.srcs)
         # Local sources are read directly from the workspace by ``bazel run``.
         # Only sources from external repositories must be staged in runfiles.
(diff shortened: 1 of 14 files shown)


AlexanderLanin:
closing in favor of #784

PR #784, opened 2026-09-01
fix: exclude nested bundle mount roots

### LINKED ISSUE #781: fix: respect Bazel package boundaries when mounting docs bundles

AlexanderLanin:
## Why this is needed

When a directory-backed documentation bundle is mounted below another Sphinx source tree, the same documents can be discovered twice: once by the containing directory walk and once by the bundle mount. That creates duplicate document names and Sphinx-Needs IDs, and makes nested module/component layouts unreliable. The ownership boundary must be based on directories rather than the manifest's current file list so files created during live preview remain discoverable.

## What this PR achieves

This fixes #781 by making each directory-mounted bundle own its physical source subtree while excluding nested bundle roots from containing walks. For directory mounts, every document is therefore discovered by exactly one Sphinx source: the primary tree or its owning bundle mount.

## Changes

- Add structural exclusions for bundle roots below the application's primary source directory.
- Add structural exclusions for child bundle roots below a parent directory mount.
- Preserve directory walking for directory mounts, including files created during live preview.
- Keep explicit `docs_bundle(srcs = [...])` mounts in file-list mode.
- Extend the existing reference integration with module/component packages nested below each module's `docs` directory, covering both legacy `data` and modern `external_needs` APIs.
- Document the source-ownership boundaries and runtime behavior.

## Known gap

Explicit `docs_bundle(srcs = [...])` mounts are intended for generated documentation sources outside the primary source tree and remain in file-list mode. They are not included in the structural directory-exclusion calculation. If a workspace source file is explicitly mounted from below the primary source tree or another directory mount, it can still be discovered twice; handling that case with exact-file exclusions is left for a follow-up.

The regression coverage verifies the directory exclusion rules directly and renders the nested reference integration, including the module-level and integration-level component paths. Existing subdirectory-bundle coverage continues to verify nested bundle composition without duplicate Needs.


src/extensions/score_mounts/__init__.py
```diff
@@ -194,13 +214,119 @@ def _configured_source_suffixes(config: Config) -> tuple[str, ...]:
...
+    """
+    _, parent_dir = source_mounts[parent_index]
+    patterns: set[str] = set()
+    for child_index, (_, child_dir) in enumerate(source_mounts):
+        if child_index == parent_index:
+            continue
+        pattern = _nested_mount_pattern(parent_dir, child_dir)
+        if pattern is not None:
+            patterns.add(pattern)
+    return sorted(patterns)
```
MaximilianSoerenPollak: Wonder if this could be done in one, so we save some iterations here.

src/extensions/score_mounts/__init__.py
```diff
@@ -245,13 +389,18 @@ def _on_config_inited(app: Sphinx, config: Config) -> None:
...
+
+        # This directory was validated during the ownership pass above. Reuse its
+        # resolved spelling so the exclusion patterns and the runtime mount refer
+        # to exactly the same physical root.
+        index, walk_dir = source_mounts_by_id[id(spec)]
+        runtime_mounts.append(
+            _make_mount_entry(
+                walk_dir,
+                spec,
+                tuple(_nested_mount_excludes(index, source_mounts)),
```
MaximilianSoerenPollak: You only use this `nested_mount_excludes` here, so why not make the return type a tuple directly, instead of a list?

src/tests/docs_bzl/scenarios/reference_integration/modern_module/docs/components/component/BUILD
```diff
@@ -5,20 +5,21 @@
 # information regarding copyright ownership.
 #
 # This program and the accompanying materials are made available under the
-# terms of the Apache License 2.0 which is available at
+# terms of the Apache License Version 2.0 which is available at
```
MaximilianSoerenPollak, line 8: Copyright changes oO. weird but alright :D

src/tests/docs_bzl/scenarios/subdirectory_bundle/producer/docs/embedded/BUILD
```diff
@@ -5,7 +5,7 @@
 # information regarding copyright ownership.
 #
 # This program and the accompanying materials are made available under the
-# terms of the Apache License Version 2.0 which is available at
+# terms of the Apache License, Version 2.0 which is available at
```
MaximilianSoerenPollak, line 8: You need to tell the clanker to leave the copyright alone :D 

It changed it again in a different way.

src/tests/docs_bzl/scenarios/reference_integration/modern_module/docs/components/component/BUILD
AlexanderLanin: separate follow-up after https://github.com/eclipse-score/tools/issues/25

src/tests/docs_bzl/scenarios/subdirectory_bundle/producer/docs/embedded/BUILD
AlexanderLanin: separate follow-up after https://github.com/eclipse-score/tools/issues/25

src/extensions/score_mounts/__init__.py
AlexanderLanin: Done. I combined the primary and nested exclusion calculation into one precomputation. The runtime mount loop now reuses the calculated results instead of rescanning the complete mount list for every parent.

src/extensions/score_mounts/__init__.py
AlexanderLanin: Done. The per-mount exclusion results are now returned as tuples directly, so the runtime code no longer needs to convert the result with tuple(...).

MaximilianSoerenPollak (approved):
My issues with the code have been addressed. 
From my side this is alright now.

PR #786, opened 2026-09-01
feat: treat generated bundle docs as sources

AlexanderLanin:
## Summary

- Add explicit `docs_bundle(srcs = [...])` support for generated documentation sources.
- Keep `docs_bundle(data = [...])` for supporting/runtime files.
- Preserve generated source roots through the bundle manifest and resolve them from `bazel-bin` under `bazel run`.
- Migrate the metamodel bundle and the generated-source integration fixture.
- Update the generated-docs and Bazel macro documentation.

## Design notes

The generated metamodel RST and Mermaid output remain together in `srcs`, because the RST references the Mermaid file by a relative path and both files form one documentation source tree. Explicit bundle sources must share one parent directory and cannot be combined with `source_dir`.

## Validation

- `pre-commit run --all-files`
- `bazel test //src/extensions/score_mounts:score_mounts_tests --noshow_progress`
- `bazel run //src/tests/docs_bzl/scenarios/data_files_runfiles:docs --noshow_progress`
- `bazel run //src/tests/docs_bzl/scenarios/nested_bundles:docs --noshow_progress`
- `bazel build //:needs_json --noshow_progress`
- `bazel run //:docs --noshow_progress`


docs.bzl
```diff
@@ -101,18 +101,34 @@ _generated_conf = rule(
     },
 )
 
-def docs_bundle(name, source_dir = None, data = [], entry_doc = "index", bundles = [], scan_code = [], code_targets = [], visibility = None, **kwargs):
+def docs_bundle(
+    name,
+    source_dir = None,
+    srcs = [],
```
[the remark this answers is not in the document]
AlexanderLanin: not relevant

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,46 @@ def _bundle_runtime_path(ctx):
...
+    consistently for the complete source set.
+    """
+    first_file = files[0]
+    # Generated outputs need their execroot path; workspace files resolve from
+    # their runfiles short path.
+    first_path = first_file.path if not first_file.is_source else first_file.short_path
+    first_is_source = first_file.is_source
+    separator = first_path.rfind("/")
+    if separator < 0:
+        fail("explicit bundle source %r has no parent directory" % first_path)
```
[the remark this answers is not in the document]
AlexanderLanin: Fixed in 82fa9c07. Root-package sources without a slash now use `.` as their parent, and subsequent source comparisons use the same representation. The relevant Bazel and end-to-end checks pass.

docs/how-to/bundles/examples.rst
```diff
@@ -51,12 +51,13 @@ If the parent is mounted at ``guides/example``, its page is rendered at
...
+Mount generated bundle content
+------------------------------
 
-When a build action produces documentation rather than a source-tree ``.rst``
-file, create a bundle without ``source_dir``. Its generated files are the
-bundle's complete payload. You can generate and mount a page like this:
+Use explicit ``srcs`` when a build action produces the documentation rather
+than a source-tree ``.rst`` file. The source-less bundle has no ``source_dir``;
+its generated files are the bundle's complete source payload. You can generate
+and mount a page like this:
```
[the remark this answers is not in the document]
AlexanderLanin: Fixed in 82fa9c07. The copyable example now uses `srcs = [":generated_page"]` and is labeled as a source bundle. The separate legacy `data` path is covered by the added compatibility test.

docs.bzl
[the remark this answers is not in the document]
AlexanderLanin: Keeping `srcs` before `data` is intentional and follows the requested public API. We accept the positional-argument compatibility trade-off; current examples use keyword arguments. Therefore I am leaving this ordering unchanged.

src/extensions/score_mounts/_resolver.py
```diff
@@ -98,7 +106,25 @@ def resolve_walk_dir(
...
+        if ws_root is not None:
+            # Generated source files are exposed through bazel-bin at runtime,
+            # while their manifest paths are execroot-relative bazel-out paths.
+            output_parts = spec.src_root.split("/")
+            if (
+                len(output_parts) >= 4
+                and output_parts[0] == "bazel-out"
+                and output_parts[2] == "bin"
+            ):
+                return ws_root / "bazel-bin" / "/".join(output_parts[3:])
```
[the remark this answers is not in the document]
AlexanderLanin: Fixed in a74be126. Generated roots ending at `bazel-out/<config>/bin` are now mapped to `<workspace>/bazel-bin` by accepting the three-component root. Added a regression test for a generated `index.rst` directly at that root; the mount resolver tests pass.

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,46 @@ def _bundle_runtime_path(ctx):
...
+    consistently for the complete source set.
+    """
+    first_file = files[0]
+    # Generated outputs need their execroot path; workspace files resolve from
+    # their runfiles short path.
+    first_path = first_file.path if not first_file.is_source else first_file.short_path
+    first_is_source = first_file.is_source
+    separator = first_path.rfind("/")
+    # A source in the root package has the workspace root as its parent.
+    runtime_path = first_path[:separator] if separator >= 0 else "."
```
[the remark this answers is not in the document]
AlexanderLanin: Correct. Fixed in be807ac0: explicit srcs are now copied by a Bazel action into a dedicated tree artifact containing only the declared files, and the manifest mounts that tree. Added an integration regression test with an undeclared sibling; the declared source is mounted while the sibling is not. The generated-source, legacy data, and score_mounts tests pass.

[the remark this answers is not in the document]
AlexanderLanin: Follow-up: I replaced the materializing implementation before updating the branch. The final fix is 4ef3b50f: it keeps the original files and uses sphinx-mounts file-list mode with the declared allowlist. The thread remains resolved.

docs.bzl
```diff
@@ -130,7 +146,15 @@ def docs_bundle(name, source_dir = None, data = [], entry_doc = "index", bundles
       **kwargs: Additional attributes forwarded to the underlying rule.
     """
 
-    srcs = glob_doc_sources(source_dir) if source_dir != None else []
+    if source_dir != None and srcs:
+        fail(
+            ("docs_bundle(%s): srcs cannot be combined with source_dir; " +
+             "put generated sources in a dedicated bundle") % name,
+        )
```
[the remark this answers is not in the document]
AlexanderLanin: Correct. The validation already existed, but it was not covered by the public docs_bzl integration suite. Added :bad_source_and_srcs and asserted the analysis diagnostic "srcs cannot be combined with source_dir". The integration test passes.

src/extensions/score_mounts/__init__.py
```diff
@@ -202,6 +216,12 @@ def _on_config_inited(app: Sphinx, config: Config) -> None:
     for spec in manifest.mounts:
         if not spec.src_root:
             continue
+        if spec.files:
+            # Explicit source bundles use sphinx-mounts' file-list mode so the
+            # original files are read directly without discovering siblings.
+            source_files = resolve_source_files(manifest, spec, ws_root, runfiles_dir)
+            runtime_mounts.append(_make_file_mount_entry(source_files, spec))
```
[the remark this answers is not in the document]
AlexanderLanin: Correct. Verified against pinned sphinx-mounts 0.1.4: passing index.rst and metamodel_classes.mmd together makes file-list mode skip the whole mount. Fixed in 1094b368: all declared files remain original Bazel inputs/runfiles, while only files matching the configured Sphinx source suffixes are registered as documents; companion assets remain available beside the mounted RST. The Mermaid asset integration test passes, and bazel run //:docs now renders reference/metamodel/index.html successfully.

src/extensions/score_mounts/__init__.py
```diff
@@ -202,6 +226,25 @@ def _on_config_inited(app: Sphinx, config: Config) -> None:
     for spec in manifest.mounts:
         if not spec.src_root:
             continue
+        if spec.files:
+            # Explicit source bundles use sphinx-mounts' file-list mode so the
+            # original files are read directly without discovering siblings.
+            source_files = resolve_source_files(manifest, spec, ws_root, runfiles_dir)
```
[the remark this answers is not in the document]
AlexanderLanin: Thanks, this is valid for the transitional compatibility implementation, but I am intentionally not fixing it in this PR. The recursive data mount exists only to preserve legacy docs_bundle(data = [...]) behavior. Once legacy data discovery is removed, explicit-source bundles will no longer receive that recursive data mount, so the srcs file allowlist cannot be bypassed. I will address this as part of removing the legacy behavior rather than add a separate temporary mounting path here.

docs/reference/bazel_macros.rst
```diff
@@ -169,16 +178,21 @@ Signature: ``docs_bundle(name, source_dir = None, data = [], entry_doc = "index"
...
-  Supporting or generated files owned by this bundle. These files are part of
-  the bundle payload and are available at the bundle's eventual mount path;
-  they are useful for generated documentation sources such as a generated
-  ``index.rst``. If ``source_dir`` is omitted and ``data`` contains the
-  bundle's deliverable, the result is a bundle containing only supporting
-  files. Use this attribute for every file that belongs with a mounted bundle.
+  Supporting or runtime files owned by this bundle. These files are part of the
+  bundle payload and are available at the bundle's eventual mount path, but are
+  not processed as the bundle's documentation sources. Use ``docs(data = [...])``
+  for supporting files in the root bundle and this attribute for child bundles.
```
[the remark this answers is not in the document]
AlexanderLanin: Agreed, the current wording describes the intended new API but not the transitional compatibility behavior. I am intentionally not changing the documentation in this PR because legacy data discovery is being retained only temporarily. When that behavior is removed, data will be strictly supporting/runtime input and srcs will be the documentation-source path, so the existing wording will become accurate again. I will update the documentation together with that legacy removal instead of documenting a behavior that is about to disappear.

AlexanderLanin:
Regarding Copilot's suppressed note at docs.bzl:125 about mixing workspace and generated srcs: I agree this identifies an implementation restriction, but I am intentionally not changing this PR. It is an uncommon edge case without a current supported use case, and adding it to the public documentation would add noise around an internal limitation. The existing analysis error already names the restriction. If mixed source kinds become a required use case, I would redesign the runtime path resolution and update the public contract then, rather than expand the documentation now.

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,62 @@ def _bundle_runtime_path(ctx):
     runfiles. Keep that spelling here; ``_bundle_execroot_path`` converts it to
     the corresponding ``external/<repo>/...`` form for build actions.
     """
-    source_file = ctx.files.srcs[0].short_path
```
MaximilianSoerenPollak, line 140: A comment here why it takes the first item would be helpful I think.

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,62 @@ def _bundle_runtime_path(ctx):
...
     if source_file.startswith("../"):
         path_parts = source_file.split("/")
         external_prefix = path_parts[0] + "/" + path_parts[1] + "/"
     return external_prefix + ctx.attr.strip_prefix.rstrip("/")
 
+def _source_target_path(source_file):
+    """Return the path spelling used by the source-target staging action."""
+    # Generated files need their execroot path; workspace files resolve from
+    # their runfiles short path.
+    return source_file.path if not source_file.is_source else source_file.short_path
```
MaximilianSoerenPollak, line 151: Comment what `is_source` means would be helpful, or is that stated somewhere else?

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,62 @@ def _bundle_runtime_path(ctx):
...
+    Source targets are either workspace files or generated outputs. Their
+    paths use different Bazel spellings, so the kind of the first file is used
+    consistently for the complete source set.
+    """
+    first_file = files[0]
+    first_path = _source_target_path(first_file)
+    first_is_source = first_file.is_source
+    separator = first_path.rfind("/")
+    # A source in the root package has the workspace root as its parent.
+    runtime_path = first_path[:separator] if separator >= 0 else "."
```
MaximilianSoerenPollak, line 168: Why is `.` the fallback?

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,62 @@ def _bundle_runtime_path(ctx):
...
+    for source_file in files[1:]:
```
MaximilianSoerenPollak, line 169: Whats special about the first source, compared to the rest? I don't quiet understand this.

bzl/bundle_rules.bzl
```diff
@@ -137,13 +135,62 @@ def _bundle_runtime_path(ctx):
...
+        # A single bundle entry cannot combine source roots from the workspace
+        # and bazel-out because they have different runtime resolution rules.
+        if source_file.is_source != first_is_source:
+            fail("explicit bundle sources cannot mix workspace and generated files")
```
MaximilianSoerenPollak: Maybe add where this error was found in the msg?

bzl/bundle_rules.bzl
```diff
@@ -245,7 +297,12 @@ def _docs_bundle_impl(ctx):
     own_external_runfiles = []
     own_data = depset(direct = ctx.files.data)
 
-    if ctx.files.srcs:
+    # The macro validates this combination before creating the rule; retain
+    # the rule-level check for callers of the internal helper as well.
+    if ctx.files.source_dir_globbed and ctx.files.source_targets:
+        fail("a bundle cannot combine source_dir sources with explicit source targets")
```
MaximilianSoerenPollak: Might be nice to tell the user which bundle it is that this error occured.

src/extensions/score_mounts/_resolver.py
```diff
@@ -98,7 +115,27 @@ def resolve_walk_dir(
...
+            if (
+                # A generated file may be directly below the configuration's
+                # ``bin`` directory, so the source root itself can end there.
+                len(output_parts) >= 3
+                and output_parts[0] == "bazel-out"
+                and output_parts[2] == "bin"
+            ):
+                return ws_root / "bazel-bin" / "/".join(output_parts[3:])
+            return ws_root / spec.src_root
+        return Path.cwd() / spec.src_root
```
MaximilianSoerenPollak, line 144: This still seems so fragile / explicit. But I can't think of a better soltuion right now either. 

Maybe add here `pottential` returns in teh comments so we know what ws_root / spec.src_root or Path.cwd() / spec.src_root or the bazel wrangling output could look like? 
or you think this is not needed?

MaximilianSoerenPollak (changes requested):
Some questions but overall this looks quiet good I would say. 

Lot's we already talked about in the call yesterday so I only commented what I now saw and noticed.

bzl/bundle_rules.bzl
AlexanderLanin: yeah it is somewhere, but I can repeat it here.

src/extensions/score_mounts/_resolver.py
AlexanderLanin: Add debug prints at least?

bzl/bundle_rules.bzl
AlexanderLanin: Addressed in 6a16e836. The comment now explains that a root-package source has no parent component and that . represents the workspace or execroot root.

bzl/bundle_rules.bzl
AlexanderLanin: Addressed in 6a16e836. The docstring now explains that one manifest entry has one runtime root and one generated/source classification, so the first file establishes the canonical representation and the remaining files are validated against it.

bzl/bundle_rules.bzl
AlexanderLanin: Addressed in 6a16e836. The mixed-source error now includes both offending source paths.

bzl/bundle_rules.bzl
AlexanderLanin: Addressed in 6a16e836. Added an explicit comment that Bazel marks workspace files with is_source; generated outputs use the execroot path while workspace files use their runfiles short path.

src/extensions/score_mounts/_resolver.py
AlexanderLanin: Addressed in 6a16e836. Added concrete bazel run and sandbox path examples to the resolver docstring. I did not add debug prints because the resolution is deterministic and the existing tests cover the path variants.

bzl/bundle_rules.bzl
AlexanderLanin: Addressed in 6a16e836. Added a comment explaining that all files are globbed from the one source_dir, so the first file is representative only for detecting an external-repository prefix.

bzl/bundle_rules.bzl
AlexanderLanin: Addressed in 6a16e836. The internal rule-level error now includes the bundle label via ctx.label.

AlexanderLanin:
Addressed the two suppressed findings from review 5086178986 in 2f93014a:

- `_ensure_unique_entries` now keys explicit entries by their declared file paths relative to the shared source root. Disjoint `srcs` bundles may share a parent; directory mounts and overlapping explicit files still fail. The new `shared_parent_bundle` analysis regression test covers this.
- `_toml_path` now classifies the lexical absolute path before resolving symlinks. This preserves `bazel-bin/...` and stable external runfile spellings, while retaining the resolved-path fallback for canonical entries. Regression tests cover both cases.

Validation: `pre-commit` passed and `bazel test --lockfile_mode=error //... --build_tests_only` passed (21/21).

These were suppressed review findings rather than inline threads, so GitHub provides no individual threads to resolve.

MaximilianSoerenPollak (approved):
Seems fine now for me. 

THe comments make it a bit clearer and hopefully will help us in the future whne we look at this again.

PR #744, opened 2026-08-22
fix: resolve #737 - missing consistency check during metamodel parsing

KIRA-L001:
Closes #737.

Generated fix for: missing consistency check during metamodel parsing

Implements a self-contained solution in `contrib_737_missing_consistency_check_during_metamod.py`.

PR #793, opened 2026-09-01
fix: reject undeclared links while parsing the metamodel

### LINKED ISSUE #737: missing consistency check during metamodel parsing
issue #725 should have been caught during metamodel parsing

SebStefenel:
## 📌 Description

Fixes #737.

### What was missing

A need type may list link names under `mandatory_links` / `optional_links`, and
those names must be declared in `needs_extra_links`. Nothing verified that.

An undeclared link fails **silently**: sphinx-needs only creates a need option
for links it knows about, so the link is never rendered, and graph checks
referring to it never fire. There is no warning at build time, so the
documentation simply comes out missing a relationship.

This already happened. `dec_rec` used `affects` as an optional link while
`needs_extra_links` did not declare it; the missing declaration was added in
#725, and #737 asked for the check that should have caught it.

### The check

`_validate_link_declarations()` runs inside `load_metamodel_data()` — i.e.
while parsing, as the issue asks. It follows the pattern already used by
`_parse_needs_types()`: collect every offending link, then raise `SystemExit`
with the full list, so one run shows all the work rather than one error at a
time.

```
ERROR: Please declare these links in 'needs_extra_links' of the metamodel.yaml,
as undeclared links are silently ignored:
Directive 'type1': mandatory_links uses 'link1', which is not declared in 'needs_extra_links'.
```

### sphinx-needs built-in links

`links` and `parent_needs` are **not** declared in our metamodel, yet `tsf`,
`tenet`, `assertion` and `std_req` all use `links`. That is legitimate:
sphinx-needs registers both itself when a configuration does not define them
(`sphinx_needs/needs.py` — *"The default link name. Must exist in all
configurations. Therefore we set it here for the user."*).

They are allow-listed in `_SPHINX_NEEDS_BUILTIN_LINKS`, with that source quoted
in the code so the reasoning does not have to be re-derived. Without it the
check reports four false positives on a clean checkout.

### The check found an existing defect immediately

`tests/model/simple_model.yaml` declared `type1` as using `link1` and `link2`,
while its `needs_extra_links` only defined `link_option1` — the exact defect
this issue is about, sitting in the test fixture. The fixture now declares both
links, and `test_load_metamodel_data` asserts them.

### Note on #744

#744 previously attempted this issue and was closed. This is an independent
implementation: the check lives inside the parser rather than in a standalone
module, handles the sphinx-needs built-in links, and comes with tests.

## 🚨 Impact Analysis

No behaviour change for a correct metamodel — `load_metamodel_data()` returns
exactly what it did before. A metamodel that uses an undeclared link, which
previously produced silently broken documentation, now fails while parsing.

The metamodel shipped with this extension already satisfies the check
(50 need types, 27 declared links, verified by a test).

## ✅ Checklist

      — no user-facing documentation needed changing: the metamodel reference
      under `docs/` is generated from `metamodel.yaml` and describes the model
      itself, not the rules for authoring it. The rule, its consequence, and
      the sphinx-needs source for the built-in links are documented in the
      function's docstring and comments.
      — six new tests in `test_metamodel_load.py`: a declared link is accepted;
      an undeclared link is rejected in `mandatory_links` and in
      `optional_links` (parametrised); the sphinx-needs built-ins are accepted;
      several offenders are reported in one run; and the shipped metamodel
      satisfies the check.

## Verification

- `bazel test //src/extensions/score_metamodel/...` → **8 / 8 pass**, including
  the file-based tests that build real documentation with the metamodel.
- `test_metamodel_load.py` → 8 passed (2 before).
- `pytest src/extensions tools` → **289 passed**, 2 failed. Both failures
  reproduce identically on unmodified `upstream/main` (verified by checking it
  out) and are unrelated to this PR.
- `basedpyright` on the changed files → `0 errors, 0 warnings`.
- `ruff check` and `ruff format --check` → clean.


src/extensions/score_metamodel/tests/test_metamodel_load.py
```diff
@@ -92,3 +101,111 @@ def test_load_metamodel_data():
...
+        load_metamodel_data(yaml_path=model)
+
+    message = str(excinfo.value)
+    assert "ghost_link" in message
+    assert "type1" in message
+    assert section in message
+
+
+def test_sphinx_needs_builtin_links_are_accepted(tmp_path: Path):
+    """'links' and 'parent_needs' are provided by sphinx-needs itself.
```
AlexanderLanin, line 171: we actually don't want to accept those. we do not use them. ideally we would reject them.

src/extensions/score_metamodel/yaml_parser.py
```diff
@@ -165,6 +165,61 @@ def _parse_links(links_dict: dict[str, dict[str, str]]) -> dict[str, dict[str, s
...
+    be declared in `needs_extra_links`. sphinx-needs only creates a need option
+    for links it knows about, so an undeclared link fails silently: the link is
+    never rendered, and graph checks referring to it never fire. Because
+    nothing reports this at build time, it has to be caught while parsing the
+    metamodel.
+
+    This is not hypothetical. `dec_rec` used `affects` as an optional link
+    while `needs_extra_links` did not declare it; the missing declaration was
+    added in #725, and this check was requested in #737 so the next occurrence
+    fails loudly instead.
```
AlexanderLanin:
```suggestion
```

src/extensions/score_metamodel/yaml_parser.py
```diff
@@ -165,6 +165,61 @@ def _parse_links(links_dict: dict[str, dict[str, str]]) -> dict[str, dict[str, s
...
+    """
+    known_links = set(needs_links) | _SPHINX_NEEDS_BUILTIN_LINKS
+
+    errors: list[str] = []
+    for directive_name, need_type in sorted(needs_types.items()):
+        # Only the "*_links_str" fields hold the link names at this point. The
+        # resolved "*_links" fields stay None until postprocess_need_links()
+        # runs, which is after parsing.
+        used_links = [
```
AlexanderLanin: this function might be easier of you drop "mandatory_links" + "optional_links"? I dont think they improve the error message.

src/extensions/score_metamodel/yaml_parser.py
```diff
@@ -165,6 +165,61 @@ def _parse_links(links_dict: dict[str, dict[str, str]]) -> dict[str, dict[str, s
...
+            if link_name not in known_links:
+                errors.append(
+                    f"Directive '{directive_name}': {section} uses "
+                    f"'{link_name}', which is not declared in 'needs_extra_links'."
+                )
+
+    if errors:
+        raise SystemExit(
+            "ERROR: Please declare these links in 'needs_extra_links' of the "
+            "metamodel.yaml, as undeclared links are silently ignored:\n"
```
AlexanderLanin: undeclared links lead tot his error 😆 

```suggestion
            "ERROR: Please declare these links in 'needs_extra_links' of the "
            "metamodel.yaml:\n"
```

AlexanderLanin (changes requested):

AlexanderLanin:
@SebStefenel thanks for the contribution, please have a look at the review comments

src/extensions/score_metamodel/yaml_parser.py
SebStefenel: Applied.

src/extensions/score_metamodel/yaml_parser.py
SebStefenel: Sounds good, dropped it.

src/extensions/score_metamodel/yaml_parser.py
SebStefenel: Applied.

src/extensions/score_metamodel/tests/test_metamodel_load.py
SebStefenel: Don't we use them though? metamodel.yaml has links: ANY under optional_links for tsf (line 71), tenet (81), assertion (91) and std_req (103), all process/standards types.

AlexanderLanin: @MaximilianSoerenPollak wtf? 😆

MaximilianSoerenPollak: As far as I just read in the code default **`options`** are removed unless they are in the `known overlaps` list. 
Therefore if you have an OPTION that is not in that list, but in the `default keys` and is an OPTION for us (mandatory or optional) then you will get a warning and be told that you have to fix this. 

For links we do NOT do any of the `removal` of defaults. Though there is **only** the `links` default for links. 

So, yes we do accept `links` as a link type implicitly as it's defined by default as a link in sphinx-needs. So theoretically you can have a `links` parameter in any need currenntly.

MaximilianSoerenPollak: The question now is, do we want to accept this one buildin one, or not.

src/extensions/score_metamodel/tests/test_metamodel_load.py
```diff
@@ -92,3 +101,110 @@ def test_load_metamodel_data():
...
+        load_metamodel_data(yaml_path=model)
+
+    message = str(excinfo.value)
+    assert "ghost_one" in message
+    assert "ghost_two" in message
+
+
+def test_shipped_metamodel_declares_every_link_it_uses():
+    """The metamodel shipped with this extension must satisfy the check."""
+    load_metamodel_data()
```
MaximilianSoerenPollak, line 210: No tests in this test? 

Or is it just the call must execute without error?

src/extensions/score_metamodel/tests/test_metamodel_load.py
AlexanderLanin: @MaximilianSoerenPollak maybe we merge this as-is and create a follow up issue?

MaximilianSoerenPollak: I think that makes sense. 
As we would explicitly make that decision and then via downstream tests we could see what actually breaks etc.

src/extensions/score_metamodel/tests/test_metamodel_load.py
AlexanderLanin: yeah thats how I understand it, as there is no return value but raise of exceptions

AlexanderLanin (approved):
merge if/when green

PR #796, opened 2026-09-02
fix: data needs_json

AlexanderLanin:
see individual commits:

1. add scenario for reproduction
2. provide fix
3. restructure scenario and make it way bigger and understandable


src/tests/docs_bzl/test_reference_integration.py
```diff
@@ -0,0 +1,70 @@
...
+        / "index.html"
+    ).is_file()
+    assert (result.build_dir / "modules" / "modern_module" / "index.html").is_file()
+    assert (
+        result.build_dir
+        / "modules"
+        / "modern_module"
+        / "components"
+        / "component"
+        / "index.html"
```
MaximilianSoerenPollak, line 69: I hate this formatting. But I guess it's alright :D

MaximilianSoerenPollak (approved):
Going commit by commit made this clearer, thanks for that. 

Overall looks like a good path forward to fix this bug but still support legacy behaviour. 

We might need ot look more internally and see if we can simplify some things there to make it easier for ourselfs.

PR #798, opened 2026-09-03
refactor: introduce wrapper for bundle declaration

AlexanderLanin:
## Why this matters

The top-level `docs()` build called the top-level `docs_bundle()` function. So one public function was based on another public function. This makes future improvements very difficult. Instead they should rely on a common private function.

## What changed

- Move the existing declaration logic into a shared private helper.
- Keep `docs_bundle()` as the public forwarding entry point.
- Have `docs()` use the shared helper directly.
- Preserve the existing macro arguments and declaration behavior.

This is a structural preparation with no user-visible behavior change. The wrapper is intentionally retained as the extension point for future bundle-only targets.

### DIFF
--- a/docs.bzl
+++ b/docs.bzl
@@ -114,7 +114,7 @@ def _is_needs_json_target(label):
     """
     return str(label).rsplit(":", 1)[-1] == "needs_json"
 
-def docs_bundle(
+def _declare_docs_bundle(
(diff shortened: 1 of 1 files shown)


MaximilianSoerenPollak (approved):
Ye that makes sense.

PR #802, opened 2026-09-03
feat!: drop deprecated scan_code parameter

AlexanderLanin:
this is simplification only, and some new tests which thanksfully pass

### DIFF
--- a/src/tests/docs_bzl/scenarios/reference_integration/modern_module/BUILD
+++ b/src/tests/docs_bzl/scenarios/reference_integration/modern_module/BUILD
@@ -21,6 +21,7 @@ docs(
     project = "S-CORE Modern Module",
     project_url = "https://example.invalid/score-modern-module",
     source_dir = "docs",
+    test_sources = ["src/tests/docs_bzl/scenarios/reference_integration"],
     external_needs = [
         "//src/tests/docs_bzl/scenarios/reference_integration/score_platform:needs_json",
         "@score_process_description//:needs_json",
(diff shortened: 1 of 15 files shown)


src/tests/docs_bzl/scenarios/reference_integration/modern_module/BUILD
```diff
@@ -21,6 +21,7 @@ docs(
     project = "S-CORE Modern Module",
     project_url = "https://example.invalid/score-modern-module",
     source_dir = "docs",
+    test_sources = ["src/tests/docs_bzl/scenarios/reference_integration"],
```
MaximilianSoerenPollak, line 24: Does this work as just a folder not a bazel target?

MaximilianSoerenPollak (approved):
Just one question but looks simple enough I would say.

MaximilianSoerenPollak:
We need to fix module_template though if we want to merge this as well.

src/tests/docs_bzl/scenarios/reference_integration/modern_module/BUILD
AlexanderLanin: `test_sources` is a list of repo-relative paths. I was not even aware we have `test_sources`

AlexanderLanin: I just checked, it has 0 users. Next candidate for removal.

AlexanderLanin:
> We need to fix module_template though if we want to merge this as well.

coincidentally already addressed in https://github.com/eclipse-score/module_template/pull/169/changes

PR #803, opened 2026-09-03
refactor: remove deprecated source code linker cache

AlexanderLanin:
## Why

PR #358 introduced the Bazel-generated `sourcelinks_json` input and deprecated the extension-generated `score_source_code_linker_cache.json` path. The deprecated path left duplicate source-linking logic in the extension and made it unclear which component owns source scanning. This change simplifies the extension around the current Bazel-generated input.

## Open question

PR #358 retained the old path for compatibility, although I cannot think of any reason to do so. As its only used internally no-one can notice whether the compatibility layer even exists. So was there a reason? Was there doubt that the bazel way is better? Is it? We can also remove the other way of parsing. But having two parsing paths increases complexity for follow-up work significantly.

## What changed

- Remove extension-side workspace scanning and generation of `score_source_code_linker_cache.json`.
- Move the source-file parser into `scripts_bazel`, where its only consumer lives.
- Resolve the pre-generated source-links input once during the combine step, supporting both `bazel run` and `needs_json`.
- Update integration fixtures and documentation for the current flow.


src/extensions/score_source_code_linker/__init__.py
```diff
@@ -269,7 +234,7 @@ def setup_combined_linker(app: Sphinx, _: BuildEnvironment):
             "Did not find combined json 'score_scl_grouped_cache.json' in _build."
             "Generating new one"
         )
-        build_and_save_combined_file(app.outdir)
+        build_and_save_combined_file(app.outdir, app)
```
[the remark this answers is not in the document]
AlexanderLanin: We investigated this behavior in detail. The stale-cache observation is technically valid, but it predates this pull request and is unrelated to removing the deprecated source-code-linker cache.

In live preview, `incremental.py` sets `skip_rescanning_via_source_code_linker=1`. The existing guards in both `setup_combined_linker()` and `setup_repo_linker()` then retain `score_scl_grouped_cache.json` and `score_repo_grouped_scl_cache.json` whenever those files already exist. Since `_build` persists between live-preview runs, changed source-links input can therefore remain unapplied.

Those guards are unchanged on the base branch. PR #803 only removes the extension-side generation/deletion of `score_source_code_linker_cache.json`; that old file was separate from both derived caches and its deletion never invalidated them. The Bazel-provided `SCORE_SOURCELINKS` flow was already in place before this PR.

So this is a real cache-lifecycle issue, but not a regression introduced by this cleanup. It would be better handled as a separate follow-up, either by removing the derived JSON handoff entirely and keeping the grouped data in memory, or by adding explicit invalidation for the derived caches.

src/extensions/score_source_code_linker/__init__.py
```diff
@@ -87,29 +81,36 @@ def get_cache_filename(build_dir: Path, filename: str) -> Path:
...
-        # just in case someone is not using the docs(sourcelinks=...) attribute.
-        # TODO: Remove this once backwards compatibility is not needed anymore.
-        source_code_links_json = get_cache_filename(
-            outdir, "score_source_code_linker_cache.json"
-        )
+    source_code_links_path = os.environ.get("SCORE_SOURCELINKS")
+    if not source_code_links_path and app is not None:
+        source_code_links_path = str(
+            getattr(app.config, "score_sourcelinks_json", "") or ""
+        ).strip()
```
MaximilianSoerenPollak, line 93: In what scenario would app be none?

MaximilianSoerenPollak (approved):
One question but rest looks like a good simplification.

src/extensions/score_source_code_linker/__init__.py
AlexanderLanin: no realistic scenario

PR #811, opened 2026-09-09
refactor: reorganize the documentation CLI

AlexanderLanin:
## Why

The documentation CLI is a distinct runtime used by every `docs()` command, but keeping it in the top-level `src` package makes its Bazel ownership and integration contract difficult to discover and extend. Giving it a dedicated package makes the CLI boundary explicit and provides focused regression coverage for the behavior that depends on Bazel-provided configuration.

## What changed

- Move the CLI implementation and dirty-build tests into `src/docs_cli`.
- Add a dedicated Bazel package with the CLI entry point, test targets, source aggregation, and package documentation.
- Update `docs.bzl` to use the packaged entry point for all interactive documentation commands.
- Add coverage for action dispatch, cache invalidation, failed builds, runfiles resolution, and mounted bundle watch directories.
- Update related documentation and test references to the new CLI location.

### DIFF
--- a/src/docs_cli/dirty_build_test.py
+++ b/src/docs_cli/dirty_build_test.py
@@ -11,22 +11,36 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
-# Unit Tests of incremental.py
-
 import json
 from pathlib import Path
 
+import pytest
 from pyfakefs.fake_filesystem import FakeFilesystem as FFS
 
-from incremental import (
-    _mounted_watch_dirs,  # pyright: ignore[reportPrivateUsage] - white-box unit test
+from src.docs_cli import cli as docs_cli
+from src.docs_cli.cli import (
     clean_builddir_if_stale,
+    mounted_watch_dirs,
     update_module_hash,
 )
 
 _BUILD = Path("/build")
 _MODULE = Path("/MODULE.bazel")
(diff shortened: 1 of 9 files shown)


src/docs_cli/dirty_build_test.py
```diff
@@ -11,22 +11,36 @@
...
+def docs_workspace(fs: FFS, monkeypatch: pytest.MonkeyPatch) -> Path:
+    """Create the minimal workspace environment used by ``cli.main``."""
+    monkeypatch.setenv("BUILD_WORKSPACE_DIRECTORY", str(_WORKSPACE))
+    monkeypatch.setenv("PACKAGE_DIR", "component")
+    monkeypatch.setenv("SOURCE_DIRECTORY", "docs")
+    monkeypatch.setenv("DATA", "[]")
+    fs.create_dir(_WORKSPACE / "component")
+    for name in ("MODULE.bazel", "MODULE.bazel.lock", "component/BUILD"):
+        fs.create_file(_WORKSPACE / name, contents="stable")
+    return _WORKSPACE
```
MaximilianSoerenPollak, line 43: Does it test the mock then or does it acutally test implementation details?

AlexanderLanin: this is a fixture for the other tests. no test here. it's just the common setup portion for all testcases.

MaximilianSoerenPollak (approved):

PR #812, opened 2026-09-09
refactor: remove sphinxdocs dependency

AlexanderLanin:
## Why

The documentation build still depends on the external `sphinxdocs` Bazel module for Needs generation, even though the surrounding bundle and CLI behavior is maintained in this repository. Removing that split ownership simplifies the build and keeps the Needs action aligned with the repository-owned bundle provider.

## What changed

- Replace the external SphinxDocs Bazel rule with a repository-owned Needs JSON action.
- Remove the `sphinxdocs` module and lockfile dependency plus obsolete adapter code.
- Extend the docs CLI to support the Bazel Needs action while preserving normal CLI behavior.
- Retain direct bundle-source projection and add focused regression coverage.


src/docs_cli/cli.py
```diff
@@ -166,20 +165,28 @@ def sphinx_arguments(ws_root: Path, package_dir: Path, build_dir: Path) -> list[
...
     generated_config = os.environ.get("SPHINX_CONFIG_FILE", "")
     if generated_config:
-        # Under ``bazel run`` this is a runfiles-relative path.  Sphinx wants
-        # the directory containing a file literally named ``conf.py``.
+        # Actions pass an execution-root path; interactive targets pass a
+        # runfiles-relative path. Sphinx needs the containing directory.
         config_file = Path(generated_config)
-        if not config_file.is_absolute():
+        if is_bazel_build:
+            config_file = config_file.absolute()
```
MaximilianSoerenPollak, line 195: You assume here if `is_bazel_build` that config_file is absolute. 

Is that assumption correct?

src/docs_cli/cli.py
```diff
@@ -166,20 +165,28 @@ def sphinx_arguments(ws_root: Path, package_dir: Path, build_dir: Path) -> list[
...
+        elif not config_file.is_absolute():
             config_file = get_runfiles_dir() / config_file
         base_arguments.extend(["-c", str(config_file.parent)])
 
     metamodel_yaml = os.environ.get("SCORE_METAMODEL_YAML", "")
     if metamodel_yaml:
         # ``docs`` passes a runfiles-relative path under ``bazel run``.  Keep
         # the workspace-relative fallback for direct invocations.
-        if not os.path.isabs(metamodel_yaml):
+        if not is_bazel_build and not os.path.isabs(metamodel_yaml):
```
MaximilianSoerenPollak, line 207: Why eactly is the negation here. 

So if it's `needs_json` you will not put the metamodel in the runfiles dir?
Will it then just be discarded? That seems strange to me.

docs.bzl
```diff
@@ -346,9 +286,9 @@ def _declare_bundle_local_needs(
         srcs,
         entry_doc,
         sourcelinks_json,
+        data = [],
```
MaximilianSoerenPollak, line 327: Config now comes via data, right?

docs.bzl
```diff
@@ -142,33 +91,37 @@ def _needs_sphinx_docs(
...
+    sphinx_build_name = _bundle_internal_target(name, "sphinx_build")
+    py_binary(
+        name = sphinx_build_name,
+        srcs = [Label("//src/docs_cli:cli.py")],
+        data = sphinx_build_data + [tool for tool in tools if tool not in sphinx_build_data],
+        deps = sphinx_build_deps,
+        visibility = ["//visibility:private"],
+        tags = ["manual"],
     )
     sphinx_docs(
```
MaximilianSoerenPollak, line 143: We still use sphinx_docs here ^^. 
Shouldn't this be removed :D

MaximilianSoerenPollak (commented):
Sone questions for sure, but the premise looks nice.

MaximilianSoerenPollak (approved):
Talked about in a call. 

Good next step for refactoring

PR #817, opened 2026-09-10
refactor: add central config class

AlexanderLanin:
almost no-op change, introducing a new class

differences:
* changes execution-context detection based on bazel variables instead of our custom action (less maintanance)
* adds workspace validation (supplied path must be valid)

### DIFF
--- a/src/docs_cli/cli.py
+++ b/src/docs_cli/cli.py
@@ -30,7 +30,8 @@ from sphinx_autobuild.__main__ import (
 )
 
 from src.extensions.score_mounts._resolver import load_mounts_manifest, resolve_walk_dir
-from src.helper_lib import Environment, find_ws_root, get_runfiles_dir
+from src.helper_lib import Environment, get_runfiles_dir
+from src.helper_lib.config import DocsCliConfig
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/docs_cli/cli.py
```diff
@@ -276,14 +280,13 @@ def main(argv: list[str] | None = None) -> int:
...
         debugpy.wait_for_client()
 
-    action = env.get("ACTION")
-    is_bazel_build = action == "build_needs_json"
-    ws_root = env.optional_path("BUILD_WORKSPACE_DIRECTORY") or Path()
+    config = DocsCliConfig.from_environment(env)
+    ws_root = config.ws_root or Path()
     # Docs source and output are resolved relative to the package where docs()
     # was called; an empty PACKAGE_DIR denotes the workspace root.
     package_dir = ws_root / (env.optional_path("PACKAGE_DIR") or Path())
```
MaximilianSoerenPollak, line 287: So `PACKAGE_DIR` is not in the `DocsCLIConfig` ? 
On purpose?

MaximilianSoerenPollak: Same with output Dir in line 292

MaximilianSoerenPollak (commented):
Just 1 question.

src/docs_cli/cli.py
AlexanderLanin: v0 of config 😝

MaximilianSoerenPollak (approved):

PR #820, opened 2026-09-14
refactor: derive bundle runtime paths from source_dir

AlexanderLanin:
## What this achieves

Passes the original `source_dir` into bundle creation so the internal bundle rule can derive the source runtime path from the owning package and repository.

## Why this is needed

The current bundle setup passes a precomputed `strip_prefix` and infers external-repository paths from the first globbed file. Keeping the source-directory value until the rule boundary makes path ownership explicit and handles package-root and nested source directories consistently.

## Changes

- Replace the internal `strip_prefix` attribute and argument with `source_dir`.
- Derive local and external runtime paths from the bundle label and `source_dir`.
- Keep explicit-source and execroot conversion behavior unchanged.


bzl/bundle_rules.bzl
```diff
@@ -453,7 +456,7 @@ def create_bundle(
         source_dir_globbed = source_dir_globbed,
         source_targets = source_targets,
         sourcelinks_json = sourcelinks_json,
-        strip_prefix = strip_prefix,
+        source_dir = source_dir if source_dir != None else "",
```
MaximilianSoerenPollak, line 459: But since you put source_dir in default above as `""` would the source_dir if in 459 not always be true then, caues it's never the default `None`?

AlexanderLanin: when a user calls `docs` or `docs_bundle` the `source_dir` defaults to `None`. This place is exactly where `None` is transformed to `""`. Maybe it makes more sense to store `None` instead of empty strings, but that would be the next refactoring starting from this point.

AlexanderLanin:
FYI stacked PRs can apparently only be force pushed, not merged with main

bzl/bundle_rules.bzl
MaximilianSoerenPollak: I mean here then you say the attribute is `""`` per default. 

Is this not an issue? 
https://github.com/eclipse-score/docs-as-code/pull/820/changes#diff-637f1bfda16cc7049dbae85157101729a0cadc42705e2acce9653e9382e9c096R425-R427

bzl/bundle_rules.bzl
```diff
@@ -421,7 +422,9 @@ _docs_bundle = rule(
         "source_dir_globbed": attr.label_list(allow_files = True),
         "source_targets": attr.label_list(allow_files = True),
         "sourcelinks_json": attr.label(allow_single_file = True),
-        "strip_prefix": attr.string(default = ""),
```
a-zw, line 424: Why remove strip_prefix?

MaximilianSoerenPollak:
> FYI stacked PRs can apparently only be force pushed, not merged with main

Yes once ,one is approved it will get `merged` onto the next PR in the stack and rebased. 
That way you ave then one `main` PR at the end that will get merged.

bzl/bundle_rules.bzl
MaximilianSoerenPollak: Strip_prefix was a missnamed argument. It never actually stripped a prefix. 
That's why now it has been removed and actuall source_dir has been added.

bzl/bundle_rules.bzl
AlexanderLanin: That default "" is exactly what is being set here

MaximilianSoerenPollak (approved):

PR #823, opened 2026-09-14
refactor: clarify runtime to execroot path conversion

AlexanderLanin:
## What this achieves

Clarifies the helper name used when translating bundle runtime paths to paths relative to Bazel's execution root.

## Why this is needed

The helper is used for both local and external bundle source paths. Its previous name, `_bundle_execroot_path`, did not describe the input transformation clearly, which made the path representation boundary harder to understand.

## Changes

- Rename `_bundle_execroot_path` to `_convert_runtime_path_to_execroot_path`.
- Update the docstring, explanatory reference, and call sites.


MaximilianSoerenPollak (dismissed):
This seems fine to me. 

As long as we are sure that bundle_execroot_path is the only thing used in these places that should exists.

MaximilianSoerenPollak (approved):

PR #824, opened 2026-09-14
refactor: expose bundle source paths in execroot form

AlexanderLanin:
## What this achieves

Passes an execution-root-relative source path from bundle analysis directly to Needs generation.

## Why this is needed

Bundle analysis already knows both the runtime and execution-root spellings. Keeping the execution-root path in the provider makes that contract explicit and removes duplicate path conversion from the Needs rule.

## Changes

- Rename the provider field `own_source_root` to `source_dir_execroot_path`.
- Compute the execution-root path once in the bundle rule and reuse it for the manifest and provider.
- Make `needs_rules.bzl` consume the provider path directly.


MaximilianSoerenPollak (approved):
Having these stacked PR'' s makes this much easier to review I gotta admit. 

Although coudl also be because these changes are small in themselfs.

MaximilianSoerenPollak (approved):

PR #825, opened 2026-09-14
refactor: remove unused explicit source metadata

AlexanderLanin:
## What this achieves

Removes obsolete explicit-source metadata from the documentation bundle provider.

## Why this is needed

The `own_source_is_explicit` field was only read by the Sphinx adapter removed in #812. No current code consumes it, so keeping it makes the provider contract misleading.

## Changes

- Remove the `own_source_is_explicit` provider field.
- Remove its initialization and propagation.
- Continue the cleanup stack after #824.


MaximilianSoerenPollak (approved):
💯 ye, remove anything we don't need.

MaximilianSoerenPollak (approved):

PR #827, opened 2026-09-15
refactor: centralize documentation path resolution

AlexanderLanin:
## Why

Documentation path handling was partly reconstructed in `cli.py` from raw environment variables. This spread the rules for direct invocations, `bazel run`, and sandboxed `bazel build` across the launcher and made it easier for callers to disagree about path bases. Centralizing the resolved paths in `DocsCliConfig` gives the launcher one context-aware source of truth while keeping execution-relative and workspace-relative paths explicit.

## What changed

- Add cached `package_dir`, `source_dir`, `output_dir`, and `source_dir_relative_to_ws` properties to `DocsCliConfig`.
- Move the environment-specific path decisions into those properties, alongside the existing execution-environment and Git-root handling.
- Update `main()` and `sphinx_arguments()` to consume the resolved config paths instead of reconstructing them from environment variables.
- Preserve the separate Bazel build output tree and the interactive cache behavior for `bazel run` and direct invocations.
- Update the CLI tests to make the direct, Bazel run, and Bazel build path contracts explicit.


MaximilianSoerenPollak (approved):
Looked through together in a call. 💯  looks good

PR #832, opened 2026-09-15
fix: improve generic Mermaid and Need template rendering

AlexanderLanin:
## Why

This PR makes the reusable documentation rendering path reliable for generated metamodel diagrams and graph-driven Sphinx-Needs templates. Without these fixes, hyphenated type names can produce invalid or ambiguous Mermaid references, nested Needs can be missed by the file-based checks, and post-template rereads can lose the Needs needed to resolve links and backlinks.

## What changed

- Mermaid diagrams now use safe internal identifiers while retaining the original type names as displayed labels. Hyphenated names receive a short stable digest so normalization does not collapse distinct names.
- The Mermaid runtime is aligned to version 11.17.2, and the corresponding golden HTML sample is updated.
- Sphinx-Needs template state and helper naming now describe the build environment and temporarily removed Needs explicitly. Link traversal supports outgoing links and backlink fields, and type selection is documented accurately.
- Post-template rereads preserve the temporarily removed Needs while rebuilding the document, then clear the temporary snapshot.
- File-based Need counting recognizes indented nested directives.
- Comments and extension documentation explain the non-obvious lifecycle and traversal behavior.

This PR contains reusable rendering and test-infrastructure fixes; it does not introduce the tool-verification report metamodel content.


src/extensions/score_metamodel/tests/test_rules_file_based.py
```diff
@@ -32,6 +33,7 @@
 
 # Relative paths of all rst files in RST_DIR
 RST_FILES = [str(f.relative_to(RST_DIR)) for f in Path(RST_DIR).rglob("*.rst")]
+_NEED_DIRECTIVE_PATTERN = re.compile(r"^\s*\.\.\s+([A-Za-z][\w-]*)::")
```
MaximilianSoerenPollak, line 36: No way this does only what it should. 

But I don't think it matters here, as this is just a display helper, and if the regex is wrong it wont break something important.

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -46,6 +53,24 @@ def _find_need(needs: dict[str, NeedItem], need_id: str) -> NeedItem | None:
     return None
 
 
+def _get_available_needs() -> dict[str, NeedItem]:
+    """Return the current Needs, including temporarily removed Needs."""
```
MaximilianSoerenPollak, line 57: I don't remember is `global` assumed in here? 
Don't you need to do: 
```
global _build_enviroment
```

Or is this only for writing it not reading?

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -60,26 +85,76 @@ class _LinkedNeeds:
...
+        source_id = _base_need_id(need_id)
+        for candidate in needs.values():
+            points_to_source = any(
+                _base_need_id(link.to_link_string()) == source_id
+                for link in candidate.get_links(link_type, as_str=False)
+            )
+            if points_to_source and candidate["id"] not in linked_ids:
+                linked.append(candidate)
+                linked_ids.add(candidate["id"])
+        return linked
```
MaximilianSoerenPollak, line 134: I wonder if you can get around this double checking if you wait for the `needs_locked` event? 
Cause as far as I know, then the needs are actually locked and everything is evaluated (so the need should be fully complete)

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -115,44 +202,56 @@ def _rerender_pages_with_complex_post_templates(
...
+    try:
+        for docname in pages_to_reread:
+            app.emit("env-purge-doc", env, docname)
+            env.clear_doc(docname)
+            app.builder.read_doc(docname)
+    finally:
+        # The exception, if any, must still propagate. Clear the process-local
+        # snapshot first so stale Needs cannot affect later rereads or builds
+        # that continue in the same Python process.
+        _temporarily_removed_needs = {}
```
MaximilianSoerenPollak, line 240: Finally will always run no? 
so this happens always, not just in an exception like the comment states.

MaximilianSoerenPollak (commented):
Some questions

src/extensions/score_sphinx_needs_templates/__init__.py
AlexanderLanin: yeah... to be fair the comment says "if any". I can improve wording!

src/extensions/score_metamodel/tests/test_rules_file_based.py
AlexanderLanin: this is our tests. so it can only break if we change the test-rst files to something incompatible.

And since this matches too much as you mentioned, there is a check for `match.group(1) in need_directive_names`

src/extensions/score_sphinx_needs_templates/__init__.py
AlexanderLanin: not needed, as its used read only here

src/extensions/score_sphinx_needs_templates/__init__.py
AlexanderLanin: I looked into the event timing. In our Sphinx-Needs version the event is called `needs-before-sealing` (there is no `needs_locked` event). It runs after link/backlink resolution, just before the Need data is locked.

That is useful for inspecting finalized links, but this extension needs to reread report documents, which removes and recreates the page’s Needs. Doing that after link resolution would add the rebuilt Needs too late for that resolution pass (and is outside the intended read phase for adding/removing Needs).

So the current `env-updated` reread is intentional: it runs after the merged Needs and injected testcase links are available, while the collection is still in its read/mutable phase. Since backlinks are not finalized yet at that point, the helper derives reverse links from outgoing links as a fallback.

MaximilianSoerenPollak (approved):
Questions answered. Looks good to me now.

PR #835, opened 2026-09-16
refactor: use execpath in genrule commands

AlexanderLanin:
## Why

In these genrules, `$(location ...)` resolves to the same paths as `$(execpath ...)`, so this is a behavior-preserving cleanup. `location` is a legacy, context-dependent path expansion; using `execpath` makes it explicit that these commands need paths usable by the build action.

## What changed

- Replaced `$(location ...)` with `$(execpath ...)` in genrules that copy Needs outputs, generate metamodel documentation, and create generated documentation fixture files.
- Added comments explaining the shell commands and path/output substitutions.


src/extensions/score_metamodel/docs/BUILD
```diff
@@ -24,7 +24,10 @@ genrule(
         "generated/index.rst",
         "generated/metamodel_classes.mmd",
     ],
-    cmd = "$(location :generate_metamodel_rst_bin) --rst-output $(location generated/index.rst) --mmd-output $(location generated/metamodel_classes.mmd) $(location //src/extensions/score_metamodel:metamodel_yaml)",
+    # Run the generator on the YAML input and tell it where to write the RST
+    # and Mermaid files. Bazel expands each $(execpath ...) to the named tool
+    # or file's path in this build action, so the command can find them.
+    cmd = "$(execpath :generate_metamodel_rst_bin) --rst-output $(execpath generated/index.rst) --mmd-output $(execpath generated/metamodel_classes.mmd) $(execpath //src/extensions/score_metamodel:metamodel_yaml)",
```
MaximilianSoerenPollak, line 30: wonder if the last one should be also `@score_docs_as_code` or if this doesn't matter here.

MaximilianSoerenPollak (approved):
Looks good, one comment but we can address that if its ap roblem in a future PR

src/extensions/score_metamodel/docs/BUILD
AlexanderLanin: seems to work 😆

PR #836, opened 2026-09-16
refactor: route Bazel docs inputs through the CLI

AlexanderLanin:
## Summary

- route Bazel-provided documentation inputs through `cli.py`
- resolve runfiles-relative paths before forwarding them as Sphinx defines
- remove duplicate manifest and source-link path resolution from extensions
- keep the Sphinx configuration boundary consistent for metamodel, sourcelinks, known-good data, and mounts
- prepare the external-needs migration by centralizing its runfiles path construction without changing the current interface

The larger `external_needs_source` label-resolution migration remains outside this change.

## Review note

The commits are intentionally ordered as incremental refactor steps. Please review them commit by commit, in order, rather than only as one combined diff.


src/docs_cli/cli.py
```diff
@@ -179,34 +198,19 @@ def sphinx_arguments(
...
-                if runfiles_dir is not None
-                else ws_root / metamodel_yaml
-            )
-        metamodel_yaml = metamodel_yaml.absolute()
+        metamodel_yaml = _resolve_runfiles_relative_path(config, metamodel_yaml)
         base_arguments.append(f"--define=score_metamodel_yaml={metamodel_yaml}")
 
+    if sourcelinks_json := env.optional_path("SCORE_SOURCELINKS"):
+        # The sandboxed Needs action and ``bazel run`` both set this env var;
+        # only the extension reads ``app.config.score_sourcelinks_json``.
```
a-zw, line 210: It is not clear to me what "sandboxed Needs action" and "the extension" is.

a-zw (approved):
I like this. It should ease the reasoning about the different invocations.

src/docs_cli/cli.py
AlexanderLanin: will improve in next PR

PR #837, opened 2026-09-16
test: cover reference integration in downstream compatibility tests

AlexanderLanin:
## Summary

- add `reference_integration` to the downstream compatibility matrix for local and remote overrides
- support consumers that declare `score_docs_as_code` in an included module file
- document the added downstream repository coverage


a-zw (approved):
LGTM

PR #839, opened 2026-09-18
feat: platform_verification_report template

antonkri:
## Why

`module_verification_report` currently mixes two different scopes in one report:
- module-scoped Components (the report's own name/purpose)
- platform-scoped Features, reached transitively through the module's components

This makes the module report noisy for anything that cares only about its own components, and there was no way to get an aggregated, platform-wide view of *all* Features in one place.

## What changed

- `src/needs_templates/module_verification_report.need`: removed the Feature-loop section (Requirements/Architecture/Inspection Statistics per Feature reached via `belongs_to`). The report now starts at `Components` and only covers the module's own components, as its name implies.
- `src/needs_templates/platform_verification_report.need` (new): the feature-scoped counterpart. Instead of resolving Features through one module's component graph, it collects every Feature in the current build via the existing `needs_of_type("feat")` render-context helper, and renders the same Requirements/Architecture/Inspection statistics per Feature that were removed from the module report (plus a Feature Overview table, analogous to the module report's Component Overview).

## Verification

- Rendered both templates standalone with `minijinja` (the engine `sphinx_needs` actually uses for `.need` templates) against a small fake Need graph (2 features across 2 components/1 module) to confirm they render without errors and produce structurally correct RST (heading underline lengths, macro expansion, table filters).
- Exercised via `reference_integration`'s `--override_module=score_docs_as_code=<local checkout>` to confirm the module resolves and the `docs()` macro picks up the templates end-to-end (`bazel build //:docs` succeeds).
- A full `bazel run //:docs` Sphinx render against the whole reference_integration graph was not completed in this environment due to devcontainer memory limits (unrelated to the template change); would appreciate a CI/reviewer run of the reference_integration docs build against this branch to confirm the final rendered pages.

## Follow-up (not in this PR)

- `reference_integration`'s `docs/verification_report/platform_verification_report.rst` was updated in that repo to use `:post_template: platform_verification_report` instead of its previous hand-written content, to actually consume this new template.
- Scoping these reports to a release milestone (`report_version`/`valid_from`) is split out into #843, stacked on this branch.


antonkri:
@MaximilianSoerenPollak , @AlexanderLanin  please review and ideally approve ;-)

AlexanderLanin:
Thanks a lot for putting this together — this is a substantial and useful change!  
One thought for reviewability: would you be open to splitting it into two PRs? The first could focus on separating the module- and platform-scoped reports, and the second could introduce the  report_version  filtering. The existing commits already seem to map quite naturally to those two pieces, so hopefully this would mostly be a matter of rearranging the work rather than redoing it.  
I think that would make both changes easier to review and reduce the risk of unintentionally mixing separate concerns.

---

We'll then somehow split the review between .py code by docs-as-code team, and .need code by process team.

antonkri:
@AlexanderLanin done, see https://github.com/eclipse-score/docs-as-code/pull/843

AlexanderLanin:
@pahmann @masc2023 @aschemmel-tech @PandaeDo @RolandJentschETAS we haven't set up codeowners yet, but in process meeting we discussed that we should apply the same codeowners as module template for the templates here.

RolandJentschETAS (approved):

antonkri:
@AlexanderLanin as Roland has approved it, could you please approve this as well?

MaximilianSoerenPollak:
@antonkri @RolandJentschETAS question here. 

The Platform scoped report. 
At which `need` or `document` should this template be used?

RolandJentschETAS:
> @antonkri @RolandJentschETAS question here.
> 
> The Platform scoped report. At which `need` or `document` should this template be used?

https://eclipse-score.github.io/module_template/main/verification_report/module_verification_report.html#doc__mod_temp_verification_report
https://eclipse-score.github.io/process_description//main/process_areas/verification/verification_workproducts.html#wp__verification_platform_ver_report

The created WP are documents.

AlexanderLanin (approved):
We need to try that out etc and adapt our tooling. But I'll merge this as-is now as it's an improvement (over having nothing), and we have way too many open PRs.

(Once tests are green)

PR #841, opened 2026-09-20
refactor: expose bundle target metadata

AlexanderLanin:
## What this achieves

Bundle metadata and direct code targets are now carried from Bazel through the
composition manifest into Python. Python can therefore receive the information
about which bundle declared a physical entry and which implementation targets
belong directly to it.

This PR only transports and parses that information. It does not yet use the
new metadata for additional runtime behavior.

## Why

Bazel already knows the declaring bundle and its direct code targets while it
composes nested and cross-repository bundles. The existing manifest discarded
that information and exposed only physical mount data. Reconstructing the
logical ownership later from paths would be unreliable, especially after
rebasing bundles into a larger documentation tree.

The metadata must therefore cross the Bazel/Python boundary before Python
features can consume it.

## Changes

- Add bundle label, bundle name, root-bundle status, and direct code-target
  metadata to the composition manifest.
- Parse the new manifest fields into Python data classes.
- Generate the manifest from the actual root bundle, while keeping the root
  entry out of the Sphinx mount list.
- Document the remaining data-based compatibility paths slated for removal.


AlexanderLanin:
those 2 are old and already fixed. copilot doesnt notice it :/

src/extensions/docs/mounts_internals.rst
```diff
@@ -37,7 +37,13 @@ Each manifest entry contains:
   walked;
 * ``mount_at`` and ``attach_to`` — the already-composed Sphinx placement; and
 * ``entry_doc`` — the canonical entry document declared by the source bundle.
-* ``external`` — whether the directory belongs to another Bazel module.
+* ``external`` — whether the directory belongs to another Bazel module;
+* ``bundle`` — the declaring bundle's Bazel label, name, and direct targets;
```
a-zw, line 41: What is the bundle metadata useful for? I can only imagine misuse, i.e. documentation logic coupled more tightly to Bazel.

I would prefer that this bundle attribute is not provided/necessary to Python code.

deprecations.md
```diff
@@ -0,0 +1,36 @@
...
+  See the NOTICE file(s) distributed with this work for additional
+  information regarding copyright ownership.
+
+  This program and the accompanying materials are made available under the
+  terms of the Apache License Version 2.0 which is available at
+  https://www.apache.org/licenses/LICENSE-2.0
+
+  SPDX-License-Identifier: Apache-2.0
+----------------------------------------------------------------------------- -->
+
```
a-zw:
```suggestion
```

We don't prefix Markdown files with this copyright header, do we?

deprecations.md
```diff
@@ -0,0 +1,36 @@
...
+# Deprecated behavior to remove
+
+This file tracks compatibility paths that should be removed after callers have
```
a-zw: A markdown file is not a good place to track deprecations. Rather, in Starlark we should print deprecation warnings.

a-zw (dismissed):

src/extensions/docs/mounts_internals.rst
AlexanderLanin: goal is to link e.g. component needs to bazel targets to source code. Visible effects will take some additional PRs.

bzl/bundle_rules.bzl
```diff
@@ -216,8 +224,11 @@ def _rebase_bundle_entry(entry, mount_at, attach_to):
     bundle's aggregate data would associate the same file with unrelated
     mounts, so the mounts resolver could select the wrong destination.
     """
-    is_bundle_root = not entry.mount_at
-    if is_bundle_root:
+    # This is the child bundle's own root before it is placed in the parent.
+    # It is distinct from ``root_bundle``, which describes ownership by the
+    # root bundle of the complete composition.
+    is_unplaced_bundle_root = not entry.mount_at
+    if is_unplaced_bundle_root:
```
MaximilianSoerenPollak: I don't quiet understand this. 

In what scenario does this occur? Is If it's not used as a `bundle`  but instead standalone correct? 
If that is the case I find the name a bit confusing, but also can't think of anything better atm.

src/extensions/docs/mounts_internals.rst
AlexanderLanin: Specifically currently there is a heuristic for how to apply code coverage to component needs. This approach will make the path deterministic / safe.

bzl/bundle_rules.bzl
```diff
@@ -236,6 +247,13 @@ def _rebase_bundle_entry(entry, mount_at, attach_to):
         # Preserve the explicit file allowlist when the entry is rebased.
         files = entry.files,
         data = entry.data,
+        # Rebasing changes only placement. Keep the declaring bundle identity
+        # and direct targets attached to the source entry as it moves through
+        # the composition graph.
+        bundle_label = entry.bundle_label,
+        bundle_name = entry.bundle_name,
+        code_targets = entry.code_targets,
+        root_bundle = False,
```
MaximilianSoerenPollak, line 261: What is the difference between `root_gundle` and `is_unplaced_bundle_root` ?

deprecations.md
AlexanderLanin: we do... hopefully this is even enforced
https://github.com/eclipse-score/tools/blob/main/cr_checker/tool/templates.ini#L79

bzl/bundle_rules.bzl
```diff
@@ -427,6 +466,9 @@ _docs_bundle = rule(
         "bundle_mount_ats": attr.string_list(),
         "bundle_attach_tos": attr.string_list(),
         "data": attr.label_list(allow_files = True),
```
MaximilianSoerenPollak, line 482: So this already has the other things? `root_bundle` label & name etc. ? Just Code targets was missing?

bzl/bundle_rules.bzl
```diff
@@ -310,6 +337,10 @@ def _docs_bundle_impl(ctx):
             # Directory mounts discover all supported files below this root.
             files = [],
             data = own_data,
+            bundle_label = own_bundle_label,
+            bundle_name = own_bundle_name,
+            root_bundle = True,
```
MaximilianSoerenPollak, line 350: How do you know that this is a root_bundle at this point?

bzl/bundle_rules.bzl
```diff
@@ -342,6 +373,10 @@ def _docs_bundle_impl(ctx):
             # original source root and therefore visits only declared files.
             files = source_files,
             data = own_data,
+            bundle_label = own_bundle_label,
+            bundle_name = own_bundle_name,
+            root_bundle = True,
```
MaximilianSoerenPollak, line 389: Same question as above

bzl/bundle_rules.bzl
```diff
@@ -363,6 +398,10 @@ def _docs_bundle_impl(ctx):
             # Pure-data entries have no documentation source allowlist.
             files = [],
             data = own_data,
+            bundle_label = own_bundle_label,
+            bundle_name = own_bundle_name,
+            root_bundle = True,
```
MaximilianSoerenPollak, line 417: I'm guessing data bundles are root_bundles by defintion?

bzl/mount_rules.bzl
```diff
@@ -16,8 +16,27 @@ Conversion of documentation bundles from Bazel into mount metadata.
...
+        target.label + "\n" + target.type
+        for target in targets
+    ])
+    return [
+        {
+            "label": value.split("\n")[0],
+            "type": value.split("\n", 1)[1],
+        }
+        for value in encoded
+    ]
```
MaximilianSoerenPollak: Why do we need to sort it?

bzl/mount_rules.bzl
```diff
@@ -16,8 +16,27 @@ Conversion of documentation bundles from Bazel into mount metadata.
...
+    pair before sorting keeps each label associated with its rule type.
+    """
+    encoded = sorted([
+        target.label + "\n" + target.type
+        for target in targets
+    ])
+    return [
+        {
+            "label": value.split("\n")[0],
+            "type": value.split("\n", 1)[1],
```
MaximilianSoerenPollak: You are only splitting the first occurence here, is this on purpose?

deprecations.md
AlexanderLanin: That's true, but its a different goal. What I wanted here is a list for us of stuff that we should remove on next major release. Let me just remove it from this PR and handle it separately.

src/extensions/score_mounts/_resolver.py
```diff
@@ -45,68 +93,65 @@ class MountSpec:
...
+        attach_to=attach_to or None,
+        entry_doc=cast("str", entry["entry_doc"]),
+        external=cast("bool", entry["external"]),
+        repository=cast("str", entry["repository"]),
+        generated=cast("bool", entry["generated"]),
+        files=cast("list[str]", entry.get("files", [])),
+        data=cast("list[str]", entry["data"]),
+        root_bundle=cast("bool", entry["root_bundle"]),
+        bundle=_read_bundle_metadata(entry),
+    )
```
MaximilianSoerenPollak: Should these not be methods of the class? Seems kinda llogical to me.

deprecations.md
MaximilianSoerenPollak: We do, since changes '& improvements in the cr_checker

deprecations.md
MaximilianSoerenPollak: I agree. 

THough I think we should have some kind of marker though that will show what is gonna be deprecated next time. 
At least in comments somewhere.

MaximilianSoerenPollak (commented):
Got some questions.

bzl/bundle_rules.bzl
AlexanderLanin: If the bundle has not been mounted anywhere, it gets is_unplaced_bundle_root. So this is basically a very dynamic attribute which can change later on.

bzl/bundle_rules.bzl
AlexanderLanin: So e.g. a child mount has is_unplaced_bundle_root = True, until it is finally mounted somewhere.

bzl/bundle_rules.bzl
AlexanderLanin: All bundles start with the assumption that they are roots, until they are placed somewhere. And _rebase_bundle_entry sets it to false.

bzl/bundle_rules.bzl
AlexanderLanin: as normal bundles, until they are wrapped somewhere they are root bundles

bzl/mount_rules.bzl
AlexanderLanin: no real reason, just for stable json output which makes it easier to use it in golden samples. will check again whether we really need it.

src/extensions/score_mounts/_resolver.py
```diff
@@ -45,68 +93,65 @@ class MountSpec:
...
+
+
+def _read_mount_spec(entry: dict[str, object]) -> MountSpec:
+    """Map one producer-owned manifest entry to a runtime mount spec."""
+    attach_to = cast("str", entry["attach_to"])
+    return MountSpec(
+        src_root=cast("str", entry["src_root"]),
+        runtime_path=cast("str", entry["runtime_path"]),
+        mount_at=cast("str", entry["mount_at"]),
```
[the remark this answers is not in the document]
AlexanderLanin: Automated response: This is addressed in the current PR. The CLI fixtures now use the same producer-shaped manifest fields as the Bazel manifest, including attach_to, entry_doc, data, root_bundle, and bundle metadata. The tests therefore exercise watch-directory behavior against the actual manifest contract.

src/extensions/score_mounts/_resolver.py
```diff
@@ -45,68 +93,65 @@ class MountSpec:
     # the mount can use the original files without recursively walking peers.
     files: list[str] = field(default_factory=list)
     data: list[str] = field(default_factory=list)
+    # Whether this physical entry belongs to the composition's root bundle.
+    # Rebasing a nested bundle changes this to false while preserving the
+    # logical bundle metadata below.
+    root_bundle: bool = False
```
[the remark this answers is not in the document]
AlexanderLanin: Automated response: This is addressed in the current PR. mounted_watch_dirs now ignores root_bundle entries because the root source is already handled by the normal Sphinx source tree. Only rebased child source and data mounts are added as additional watch paths, and the fixture covers the root-entry case.

src/extensions/docs/mounts_internals.rst
AlexanderLanin: Automated response: The metadata is needed to connect a document or Need to the Bazel bundle that supplied it and to that bundle's direct implementation targets. The current code uses a path-based heuristic for applying code coverage to component Needs; carrying this association from Bazel into Python makes that lookup explicit and deterministic. Python receives the already-resolved metadata and does not reconstruct Bazel labels or call Bazel APIs. This PR only transports the information; later processing will consume it.

deprecations.md
AlexanderLanin: Automated response: The file was removed from this PR, so the Markdown copyright-header question no longer applies.

deprecations.md
AlexanderLanin: Automated response: Agreed for this PR. deprecations.md was removed. The compatibility code keeps a local TODO next to the behavior so the future cleanup remains visible, but this PR does not introduce a separate deprecation-warning mechanism.

bzl/bundle_rules.bzl
AlexanderLanin: Automated response: The original variable was misleading and has been removed. The code now checks if not entry.mount_at directly and explains both branches. This path runs when a parent rebases entries from a child bundle; it is not specifically a standalone-versus-bundled mode.

bzl/bundle_rules.bzl
AlexanderLanin: Automated response: root_bundle is the persisted flag on the final manifest entry. It says whether the entry belongs to the root bundle of the current composition. The old is_unplaced_bundle_root expression was only a temporary condition while rebasing and has now been removed. A bundle can therefore be a root in a standalone composition and a child when embedded.

bzl/bundle_rules.bzl
AlexanderLanin: Automated response: These fields are introduced as one manifest contract. The previous manifest did not carry the bundle label, bundle name, root status, or direct code targets. code_targets is the new payload, while label, name, and root_bundle identify which bundle the targets belong to and how the entry participates in the current composition.

bzl/bundle_rules.bzl
AlexanderLanin: Automated response: At this point _docs_bundle_impl is creating an entry directly from the current bundle's own source or data. Direct entries are root entries of the current composition and are therefore marked true. If a parent embeds the bundle, _rebase_bundle_entry creates the composed entry and sets root_bundle to false. Root status is composition-specific, not a permanent property of the bundle.

bzl/bundle_rules.bzl
AlexanderLanin: Automated response: The same explanation applies here: this is the current bundle's direct explicit-source entry, so it starts as a root entry of the current composition. When a parent embeds it, _rebase_bundle_entry marks the resulting child entry as false.

bzl/bundle_rules.bzl
AlexanderLanin: Automated response: A pure-data bundle gets a root entry when considered as a standalone composition because that entry is needed to carry the bundle's data. Root here means root entry of the current composition, not a documentation source. When the bundle is nested, rebasing changes the entry to root_bundle = false so its data can be mounted as child data.

bzl/mount_rules.bzl
AlexanderLanin: Automated response: No, sorting was not required. The sorting helper has been removed. code_targets are now emitted in the producer's declaration order; consumers treat them as a collection and do not assign meaning to that order.

bzl/mount_rules.bzl
AlexanderLanin: Automated response: This is no longer relevant because the sorting helper and its artificial label/type newline encoding were removed. The current code does not split the combined string.

src/extensions/score_mounts/_resolver.py
AlexanderLanin: Automated response: Addressed. Parsing now lives in from_manifest_entry class methods on BazelTarget, BundleMetadata, and MountSpec. load_mounts_manifest delegates to MountSpec.from_manifest_entry, which composes the nested metadata objects.

antonkri (dismissed):
I think, we should talk about the overall concept first. What is a bundle? How does it correspond to the s-core metamodel? Is it a unit, a component or a module? Can we have compositions of bundles? What is the overall goal? Which use-cases we cover with this?

AlexanderLanin:
will call and in short: a bundle has no defined scope

MaximilianSoerenPollak (approved):
From my side all questions are answered and make sense.

PR #843, opened 2026-09-21
feat: scope verification reports to a report_version via valid_from

antonkri:
Stacked on #839 (targets that branch, not `main`, so the diff here only shows the two commits below; it will show against `main` once #839 merges and this PR's base is retargeted).

## Why

`module_verification_report` and `platform_verification_report` (see #839) always show every Requirement/Feature/Component in the current build. There is no way to scope either report to "everything relevant for release X", even though the metamodel already has a `valid_from`/`valid_until` attribute on `feat_req`/`stkh_req` for exactly this purpose (see `tool_req__docs_req_attr_validity_correctness`).

## What changed

- `src/extensions/score_metamodel/metamodel.yaml`: new optional `report_version` attribute on the `document` need type, using the same `vMAJOR.MINOR[.PATCH]` format already enforced for `valid_from`/`valid_until`.
- `src/extensions/score_sphinx_needs_templates/__init__.py`: two new render-context helpers.
  - `req_in_scope(need, report_version)` decides whether a single requirement Need is in scope. `feat_req`/`stkh_req` carry `valid_from` directly; `comp_req` has no `valid_from` of its own, so it inherits scope from the `feat_req` it is `derived_from`. A requirement with no resolvable `valid_from` is excluded whenever a `report_version` is set.
  - `any_req_in_scope(reqs, report_version)` decides whether a Feature/Component has *any* requirement in scope.
- `src/needs_templates/module_verification_report.need` and `src/needs_templates/platform_verification_report.need` use these helpers to:
  - restrict the Requirements Statistics needpies and the requirements needtable to `id in [...]` the in-scope requirement IDs, and
  - drop a Feature/Component from the report entirely (Overview table and its whole detail section) when none of its requirements are in scope, instead of rendering an empty section.
- An unset/empty `report_version` keeps a report fully unscoped (all Features/Components/requirements shown), which is what `doc__platform_verification_report_latest` and similar "latest" reports rely on — this is opt-in and backward compatible.


antonkri:
@AlexanderLanin as requested, please review

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -262,6 +328,10 @@ def setup(app: Sphinx) -> dict[str, object]:
     )
     app.config.needs_render_context.setdefault("linked_needs", _linked_needs_callable)
     app.config.needs_render_context.setdefault("needs_of_type", _needs_of_type_callable)
+    app.config.needs_render_context.setdefault("req_in_scope", _req_in_scope_callable)
```
AlexanderLanin: req_in_scope does not really imply "version". Maybe "req_in_report_version"?

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -262,6 +328,10 @@ def setup(app: Sphinx) -> dict[str, object]:
...
+    app.config.needs_render_context.setdefault(
+        "any_req_in_scope", _any_req_in_scope_callable
```
AlexanderLanin: same here. Like `any_req_in_report_version`. Or something prettier.

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -178,6 +178,72 @@ def __call__(self, need_type: str) -> list[NeedItem]:
...
+    def __call__(self, need: NeedItem, report_version: str | None) -> bool:
+        if not report_version:
+            return True
+
+        valid_from = need.get("valid_from")
+        if valid_from:
+            try:
+                return _parse_version(valid_from) <= _parse_version(report_version)
+            except ValueError:
+                return False
```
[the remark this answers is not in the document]
antonkri: Good catch, and you're right that it's inconsistent with the documented `valid_from`/`valid_until` semantics (`tool_req__docs_req_attr_validity_correctness`: "from" inclusive, "until" exclusive).

This is intentional scope for this PR, though: `report_version` here answers "everything relevant as of release X" (a snapshot of what should exist by then), not "everything still valid in release X" (which would additionally need the `valid_until` exclusion you describe). We deliberately kept it to the simpler `valid_from`-only check for the initial version of this feature - happy to open a follow-up to add `report_version < valid_until` support (with `comp_req` inheriting the full interval via `derived_from`, as you suggest) once there's a concrete need for it, rather than speculatively building it now.

Added focused unit tests for the current behavior (including the `comp_req`/`derived_from` inheritance path and malformed-value handling) in b71c7702.

src/extensions/score_sphinx_needs_templates/__init__.py
antonkri: Good point, renamed to `req_in_report_version` in 0cd3801d.

src/extensions/score_sphinx_needs_templates/__init__.py
antonkri: Renamed to `any_req_in_report_version` in 0cd3801d, thanks.

AlexanderLanin:
LGTM, but I would like to merge #841 first

antonkri:
> #841

Hey @AlexanderLanin , why should #841 first? From my point of view there are no correlations.

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -178,6 +178,72 @@ def __call__(self, need_type: str) -> list[NeedItem]:
...
```
MaximilianSoerenPollak, line 219: In my opinion this can never happen, or am I thinking wrong?

I do like defensive programming, but this seems a bit weird.  
If anything this defence may should be in the parser.

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -178,6 +178,72 @@ def __call__(self, need_type: str) -> list[NeedItem]:
...
+
+        linked_feat_reqs = _linked_needs_callable(need["id"], "derived_from")
+        return any(self(feat_req, report_version) for feat_req in linked_feat_reqs)
+
+
+_req_in_report_version_callable = _RequirementInReportVersion()
```
MaximilianSoerenPollak, line 225: Single variable indirections are useless imo.
Makes it harder to read.

You are only using htem on 2 places, so this makes no sense to put into a variable that is also the same lenght...

src/extensions/score_sphinx_needs_templates/__init__.py
```diff
@@ -178,6 +178,72 @@ def __call__(self, need_type: str) -> list[NeedItem]:
...
+    ``report_version`` is set and none of its requirements qualify, instead of
+    rendering an empty section. An empty/``None`` ``report_version`` always
+    returns ``True`` (unscoped reports keep every Feature/Component, even
+    ones without any requirement at all, as before).
+    """
+
+    def __call__(self, reqs: list[NeedItem], report_version: str | None) -> bool:
+        if not report_version:
+            return True
+        return any(_req_in_report_version_callable(req, report_version) for req in reqs)
```
MaximilianSoerenPollak, line 241: Why are these classes?

~~It feels a bit weird in my opinion. I would prefer functions here, these seem to not take any advantage of much a class has to offer.~~
Edit: I see that the file has set this precedent already... this might need a refactor then someday in the future.
So I guess this is okay.

AlexanderLanin:
@copilot resolve the merge conflicts in this pull request. This PR adds report_version support to the templates.

MaximilianSoerenPollak (commented):
Some comments. 

I think this is a good idea, just had some questions.

src/extensions/score_sphinx_needs_templates/__init__.py
AlexanderLanin: @MaximilianSoerenPollak sphinx needs requires this to be a class. One of the other classes explains why :(

MaximilianSoerenPollak: I must have overread that description. 
I guess then it is fine.

AlexanderLanin (approved):

PR #845, opened 2026-09-21
refactor: retain bundle metadata for documents

AlexanderLanin:
## Why

Sphinx discovers document names independently of the bundle metadata that describes their source. We need that document-to-bundle association available in Python so later processing can relate document-level data to the corresponding bundle. PR #841 provides the metadata; this stacked PR carries the association through the documentation build without consuming it yet.

## What

- Preserve the bundle-only manifest for standalone Needs actions.
- Map mounted document names to their bundle metadata and associate otherwise-discovered root documents with the root bundle.
- Keep pure data mounts out of the document mapping.
- Expose the mapping to later consumers and cover the behavior with focused tests and documentation.


MaximilianSoerenPollak:
Closed on purpose?

